PR #21, opened 2021-08-21
Add seed for assemblyscript random generator

ttulka:
Consider similar for other languages.

### DIFF
--- a/runtimes/web/src/runtime.js
+++ b/runtimes/web/src/runtime.js
@@ -128,7 +128,7 @@ export class Runtime {
             },
 
             // seed for AssemblyScript random generator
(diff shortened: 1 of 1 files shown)


aduros:
I haven't tested this yet, but shouldn't it return Date.now() (a number)?

https://www.assemblyscript.org/stdlib/math.html#using-nativemath

ttulka:
Well reviewed! You're right and should not code on Saturdays... :-D
Fixed.

aduros:
Thanks!

Let's leave this as an undocumented AS implementation detail for now until we figure out the best way to handle other languages.

PR #23, opened 2021-08-22
Add screen size const

ttulka:
What do you think about adding this constant? 

Would be handy for all games I guess...


aduros:
Sounds good to me, thanks!

PR #27, opened 2021-08-23
Snake example

ttulka:
A review is welcome.


aduros:
The code is very clean and easy to follow. I couldn't find anything to nitpick in review. It's a great example project, thanks! Merging :tada:

As a player I had a minor feedback for your consideration. It's possible to suicide by pressing the arrow backwards to the snake's direction, which can be easy to accidentally press with quick turns. It's not a huge deal, but just in case you weren't aware :)

ttulka:
Haha, very well!

Yes, I coded it so intentionally, because it is how it worked when I played Snake as a kid - I think it should be so as this retro behavior keeps the player aware :-D

aduros:
Good point :space_invader:

PR #38, opened 2021-08-30
Change stride type to u32 in Rust template blit_sub function

claudiomattera:
Parameter stride is a width, it cannot be negative.


aduros:
Thank you! :tada:

PR #41, opened 2021-08-31
Check if exists directory before bundling

### LINKED ISSUE #40: `w4 bundle` should create directory recursively if the directory not exists
Example:
(Note: `dist` currently does not exists)
```sh
$ w4 bundle build/cart.wasm --html dist/index.html
OK! Bundled dist/index.html.
```
It's better to check if `dist` exists (and create if it does not) than just attempting to write the file to the HTML file.

takaeri:
This PR _should_ fix #40. Let me know if it doesn't or have any bugs

### DIFF
--- a/cli/lib/new.js
+++ b/cli/lib/new.js
@@ -47,6 +47,10 @@ async function run (destDir, opts) {
         lang = LANGS.ASSEMBLYSCRIPT;
     }
 
+    if (!fs.existsSync(destDir)) {
+        fs.mkdirSync(destDir, { recursive: true });
+    }
+
     const srcDir = path.resolve(__dirname+"/../assets/templates/"+lang);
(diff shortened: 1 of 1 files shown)


aduros:
Hi, thanks!

This is for `w4 bundle` right? The check looks good to me, it just needs to be moved to `cli/lib/bundle.js` instead of `cli/lib/new.js`.

```js
const destDir = path.dirname(outFile);
... your existsSync check here
```

aduros:
By the way, you can test your changes by running `npm link -g` inside the cli directory. This will install a `w4` command that uses your local changes :smiley:

takaeri:
Ah yes. My bad for the wrong file. I'll submit again tomorrow.

takaeri:
Closing in favor of #45

Issue #39, opened 2021-08-30
Look into using wee_alloc by default for Rust


aduros:
https://github.com/rustwasm/wee_alloc

Issue #37, opened 2021-08-30
[Rust] cannot allocate a String in update


FaberVitale:
Hi,

I'm having trouble allocating a `String`, am I missing something?

```rust
// lib.rs
#[no_mangle]
fn update() {
    let mut my_string = String::new();

    my_string.push('#');

    text(&my_string.clone(), 20, 20);
}
```

```
Uncaught (in promise) RuntimeError: unreachable
    at update (<anonymous>:wasm-function[11]:0x20fb)
    at p.update (wasm4.js:1)
    at t (wasm4.js:1)
    at wasm4.js:1
    ```

aduros:
Possibly the same as #36

Issue #33, opened 2021-08-30
Go target.json example needs global base setting


jimmycartrette:
Starting playing off the go hello world, noticed my KEYS and Framebuffer were getting overwritten with variables in memory. Looks like this link from the rust linker options
        "--global-base=6560"
 in target.json in the go hello world cleared things up

jimmycartrette:
#34 
Created my first ever PR

jimmycartrette:
That didn't fix it, but I'm not sure how to. I haven't used a debugger but it seems like doing things that generate variables in the heap seems to put them in lower memory first, overwriting controls and videobuffer. For example this code below which just assigns a pointer to a new variable causes the keys to be pressed and a moving dot around the framebuffer

var messy *[]byte
...
	var createvar []byte = fullland
	messy = &createvar

[screenshot of corruption](https://imgur.com/wAOFkGM)

Without that, the program is completely normal. Or if I use a text() is also starts to corrupt pretty much immediately.

aduros:
Hi, thanks for reporting this!

Yeah, from what I could tell `--global-base` doesn't seem to do anything (https://github.com/rust-lang/rust/issues/46645#issuecomment-423912553)

The workaround was setting the stack size to offset the memory to the correct place... but I think it needs to be larger than 6560, to actually leave room for a stack.

Can you try changing `-zstack-size` in target.json from 6560 to 7584? (or higher) That will leave 1024+ bytes for the actual stack.

jimmycartrette:
I tried even 30336, verified that's the first stack address but the problem doesn't change. I tried to debug but it's a bit over my head. I made a repo with the first commit with no problems, and a second commit with three lines (global pointer, int and value assignment to global pointer) that causes the issue. No issues
https://github.com/jimmycartrette/LakeShooter/commit/ca9676d95c6eb8f65bec4334503cc1b87c2ac2a5

And has corruption issues:
https://github.com/jimmycartrette/LakeShooter/commit/206101cff2f7322a4a6ba488e3178ea19ace6647

Interesting though, just introducing a single global pointer variable doubles the size of my compiled code.

aduros:
Thanks, it looks like any heap allocation causes memory corruption. Here's a minimal repro case:

```go
package main

var test []byte;

//go:export update
func update () {
    test = make([]byte, 666);
}
```

[...]

PR #48, opened 2021-09-01
Add pong example

FaberVitale:
### Description

Adds to examples a simple pong game written in rust.

Nice project @aduros,  I had a ton of fun writing the cart.
AI is somewhat basic but I think it is fine overall.

### No allocation

I've tried to use wee_alloc (see #37 #39) but I had no luck .


aduros:
Nice, first Rust example! It looks excellent, as a Rust newbie, just looking through the code has taught me a few things about how to organize a game.

How was your experience overall? Besides heap allocation being busted at the moment, did you run into any other snags or things that seemed out of place from a Rust point of view?

I also have had no luck with wee_alloc so far... but I've been exploring different custom allocators designed for embedded Rust. Maybe something will work out there.

By the way, I was wondering about the cart file size. After building with `#![no_std]` and running it through `wasm-opt` from binaryen, it shrinks to a tiny 6.8 KB!

Thanks for the awesome example! :tada:

examples/pong/src/pong.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    fn tick(&mut self, gamepad: u8) {
+        self.update_p1_racket_pos(gamepad);
+        self.update_p2_racket_pos();
+        let collision = self.update_ball_pos();
+
+        if collision.intersects(CollisionType::P1_RACKET) {
+            self.play_tone(240, 3, 1, TONE_TRIANGLE);
+        } else if collision.intersects(CollisionType::P2_RACKET) {
+            self.play_tone(262, 3, 1, TONE_TRIANGLE);
```
aduros, line 283: Should the volume here be higher? It's almost muted at 1%

FaberVitale: It sounded ok during dev, but you're right it too low.

I'll submit a PR.

FaberVitale:
> How was your experience overall? Besides heap allocation being busted at the moment, did you run into any other snags or things that seemed out of place from a Rust point of view?

1.  `text` produces weird artifacts at the bottom of the screen when writing numbers: https://github.com/aduros/wasm4/blob/main/examples/pong/src/pong.rs#L355
 You do not see them because I'm painting over them: https://github.com/aduros/wasm4/blob/main/examples/pong/src/pong.rs#L367
 
2. I couldn't use trace.

3. If possible it would be better to confine all the unsafe use inside wasm4.
    Instead of giving access to raw pointers  you could wrap them in accessor and mutator functions.

4. Publish `wasm4.rs` as crate:
    this is not strictly a rust thing, it would be beneficial to have the glue code versioned and easy available through package managers.
 
---

> I also have had no luck with wee_alloc so far... but I've been exploring different custom allocators designed for embedded Rust. Maybe something will work out there.

I'm sure will find something.

In the meantime if you need dynamic arrays or strings you could use[arrayvec](https://docs.rs/arrayvec/0.7.1/arrayvec/) instead:

provides fixed size versions of `Vec` and `String`.

---

> By the way, I was wondering about the cart file size. After building with `#![no_std]` and running it through `wasm-opt` from binaryen, it shrinks to a tiny 6.8 KB!
> 
> Thanks for the awesome example! 🎉

I've played a bit with the  too but the cartridge was 64k so I didn't bother to implement it.

Few observations

1. using `no_std` and an allocator is cumbersome and [requires a nightly version of rust](https://dev.to/thepuzzlemaker/nostd-with-wasi-is-more-complicated-than-i-thought-it-would-be-14j7)

2. You cannot use some data structures like hasmap

3. [Wasm bindgen default template uses std](https://github.com/rustwasm/rust-webpack-template/blob/master/template/src/lib.rs)

See also: https://docs.rust-embedded.org/book/intro/no-std.html


I'll probably submit a PR to add no_std, it is a big size reduction and I'm not using any particular std code.

takaeri:
> 2. I couldn't use trace.

What's not working? Is there any error or just `trace` is not being executed?

FaberVitale:
@takaeri 
Ok, my bad it seems that works now, seel https://github.com/aduros/wasm4/pull/48#issuecomment-911568396

@aduros  that issue that I mentioned [above](https://github.com/aduros/wasm4/pull/48#issuecomment-911566881) happens  with trace too.

Screenshot 2021-09-02 at 14 14 54

### Code

```rust
fn num_as_str<T: Into<usize>>(score: T) -> &'static str {
    let idx: usize = score.into();

    match idx {
        10 => "10",
        _ => &DIGITS[idx..idx + 1],
    }
}

trace(num_as_str(0u8)); 
```

It logs correctly `0` but produces a weird artifact.

aduros:
Can you try increasing the stack size argument in the project's `.cargo/config.toml`? See https://github.com/aduros/wasm4/issues/33#issuecomment-908391421

FaberVitale:
> Can you try increasing the stack size argument in the project's `.cargo/config.toml`? See [#33 (comment)](https://github.com/aduros/wasm4/issues/33#issuecomment-908391421)

```diff
     "-C", "link-arg=--import-memory",
     "-C", "link-arg=--initial-memory=65536",
     "-C", "link-arg=--max-memory=65536",
-    "-C", "link-arg=-zstack-size=6560",
+    "-C", "link-arg=-zstack-size=7584",
 ```

Yep, it seems that this change fixes the issue:
I can't reproduce those artifacts using either `text` or `trace`.

PR #53, opened 2021-09-02
[png2src] Improvements, support custom templates

eXponenta:
Png2src is good, but code had a hardcoded switch-bassed generators for specific language.

I reformat code to string-tempates, now very easy add another source generators.

Other feature - custom templates, you can generate a game-specific source, for example:

```
export const %name% = {
	width: %width%,
	height: %height%,
	flags: %flags%,
	data: memory.data<u8>([ %bytes% ])
};
```

This template will emit a object export for AssemblyScript instead of constants, simmilar Trait/Object/Struct can be generated and for Rust/Go/C.

Guide for custom template is added also.

PS:
i use a template for AS like this:
```
import { Frame } from "../nodes/sprite";

const %name% = memory.data<u8>([ %bytes% ]);
export const %name%Frame = new Frame(%width%, %height%, %name%, %flags%);
```

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -34,6 +34,7 @@ program.command("png2src <images...>")
     .option("--c", "Generate C/C++ source")
     .option("--rs, --rust", "Generate Rust source")
     .option("--go", "Generate Go source")
+    .option("--t, --template <file>", "Generate source follow template", "")
     .action((images, opts) => {
         const png2src = require("./lib/png2src");
         png2src.runAll(images, opts);
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -34,7 +34,7 @@ program.command("png2src <images...>")
     .option("--c", "Generate C/C++ source")
     .option("--rs, --rust", "Generate Rust source")
     .option("--go", "Generate Go source")
-    .option("--t, --template <file>", "Generate source follow template", "")
+    .option("--t, --template <file>", "Template file with a custom output format")
     .action((images, opts) => {
         const png2src = require("./lib/png2src");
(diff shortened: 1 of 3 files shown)


cli/lib/png2src.js
```diff
@@ -2,7 +2,40 @@ const fs = require("fs");
...
+#define %name%Height %height%
+#define %name%Flags %flagsHumanReadable%
+const char %name%[%length%] = { %bytes% };
+`,
+
+    rust: 
+`const %idiomaticName%_WIDTH = %width%;
+const %idiomaticName%_HEIGHT = %height%;
+const %idiomaticName%_FLAGS = %flags%; // %flagsHumanReadable%
+const %idiomaticName%:[u8;%length%] = [ %bytes% ];
```
aduros, line 25:
```suggestion
const %idiomaticName%: [u8; %length%] = [ %bytes% ];
```

cli/cli.js
```diff
@@ -34,6 +34,7 @@ program.command("png2src <images...>")
     .option("--c", "Generate C/C++ source")
     .option("--rs, --rust", "Generate Rust source")
     .option("--go", "Generate Go source")
+    .option("--t, --template <file>", "Generate source follow template", "")
```
aduros, line 37:
```suggestion
    .option("--t, --template <file>", "Custom output format template")
```

aduros:
Nice feature! I added some small suggested tweaks, but it looks solid :smile: Will merge when I get a chance to test it.

aduros:
Thank you :tada:

PR #58, opened 2021-09-04
[refactor]  overhaul `bundle` & `run` cli commands

FaberVitale:
### Description

This is an ongoing effort to overhaul `run` and  `bundle` commands.

### Goals

- `run` & `bundle` should reuse the same config provided by `runtimes/web/webpack.config.js`.
- add more configuration options to `bundle`.
- other? Feedback and suggestions are welcomed!

To do

- [x] refactor: ovehaul `runtimes/web/webpack.config.js`
- [ ]  refactor: Use `runtimes/web/webpack.config.js` in `cli/lib/bundle.js`.
- [ ] refactor: Use `runtimes/web/webpack.config.js` in `cli/lib/server.js`.
- [ ] refactor: Move z85 encoder & decoder to its own module.
- [ ] feat: Add `bundle` options `description` `title` `favicon`.


aduros:
I think we should avoid adding webpack as a dependency to the CLI. Webpack is a necessary evil for building the JS runtime, but I'd like to keep the CLI as small and dependency-free as possible.

Adding  `w4 bundle` options like `description`, `title`, etc. sounds great to me, but I think that can be done with the simple string replacement rather than refactoring to bring in webpack.

FaberVitale:
> I think we should avoid adding webpack as a dependency to the CLI. Webpack is a necessary evil for building the JS runtime, but I'd like to keep the CLI as small and dependency-free as possible.
> 
> Adding `w4 bundle` options like `description`, `title`, etc. sounds great to me, but I think that can be done with the simple string replacement rather than refactoring to bring in webpack.
Fair enough

I'd like to add configurable description, title & favicon but manipulating html with regex can be really problematic.
I think that part really could use a proper html parser and manipulation library like [cheerio](https://www.npmjs.com/package/cheerio).

Is cheerio ok? Alternatives?

aduros:
I wonder if it wouldn't be better to include a separate html file for the bundle template. It could be rendered using something like [mustache](https://www.npmjs.com/package/mustache).

That might involve duplicating some parts of index.html, but I feel like it would be much simpler... As an extra, maybe we can move the gamepad divs and other elements that are currently in index.html into the runtime JS to remove duplication.

About the favicon, is it possible to base64 an image data URL to use as the favicon? That would be pretty cool if so!

PR #65, opened 2021-09-05
Fix typo in go setup.

0xE282B0:


aduros:
Nice catch, thank you!

PR #71, opened 2021-09-08
[perf] reduce pong.wasm cartridge size 

FaberVitale:
Notable changes

1. cart has been compiled with `no_std` and  then has been optimized using `wasm-opt`: from 64KB to 8KB.
2. updated wasm4.rs glue code.
3. added error handling with trace.
4. added brief note in readme about using `wasm-opt`.


aduros:
Thanks!

PR #75, opened 2021-09-09
Fixed small error in png2src

christopher-kleine:
In templates, "width" got replaced by "height" instead of the width.


aduros:
Oof, thanks!

PR #76, opened 2021-09-10
Changed Go template

christopher-kleine:
I changed the go template to be more Go-Idiomatic. The file `wasm4.go` was moved into it's own package.
This way the source can be more complex and it's easy to refer to it using
```go
import "cart/src/w4"

// ...

func update() {
    w4.Rect(0, 0, 10, 10)
}
````

The exported WASM is not in danger, since the `// go:export [NAME]` takes care of that.
I tested this and it works fine.

I also updated the docs to be more idiomatic. That included removing semicolons, brackets around `if` among other things.

I'd actually wanted to make it even more idiomatic by removing the source files outside of the `src`-folder.
This however is currently not as trivial since the CLI requires this folder.

### CODE REVIEW

**Comment by eXponenta** [comment]
> This however is currently not as trivial since the CLI requires this folder.

https://github.com/aduros/wasm4/blob/main/cli/lib/watch.js#L53

You can pass watchable dir as argument to watcher and use a src by default

Anyway for me built-in watch doesn't working on win10 and I don know before that it required a src folder strictly :)

File: cli/assets/templates/go/src/w4/wasm4.go [hunk 1]
```diff
@@ -47,34 +45,34 @@ const BUTTON_DOWN byte = 128;
...
 //go:export blit
-func blit (sprite *byte, x int, y int, width uint, height uint, flags uint);
+func Blit(sprite *byte, x int, y int, width uint, height uint, flags uint)
 
 /** Copies a subregion within a larger sprite atlas to the framebuffer. */
 //go:export blitSub
-func blitSub (sprite *byte, x int, y int, width uint, height uint,
-    srcX uint, srcY uint, stride int, flags uint);
+func BlitSub(sprite *byte, x int, y int, width uint, height uint,
+	srcX uint, srcY uint, stride int, flags uint)
```
aduros, line 53: Let's stick with 4 spaces indentation:

```suggestion
    srcX uint, srcY uint, stride int, flags uint)
```

site/docs/guides/user-input.md
```diff
@@ -40,10 +40,10 @@ if gamepad & BUTTON_RIGHT != 0 {
 ```
 
 ```go
-var gamepad = *GAMEPAD1;
+var gamepad = *w4.GAMEPAD1
 
-if (gamepad & BUTTON_RIGHT != 0) {
-    trace("Right button is down!");
+if gamepad&w4.BUTTON_RIGHT != 0 {
```
aduros, line 45:
```suggestion
if gamepad & w4.BUTTON_RIGHT != 0 {
```

site/docs/guides/user-input.md
```diff
@@ -110,18 +110,18 @@ void update () {
...
-    var gamepad = *GAMEPAD1;
+    var gamepad = *w4.GAMEPAD1;
 
     // Only the buttons that were pressed down this frame
     var pressedThisFrame = gamepad & (gamepad ^ previousGamepad);
     previousGamepad = gamepad;
 
-    if (pressedThisFrame & BUTTON_RIGHT != 0) {
-        trace("Right button was just pressed!");
+    if pressedThisFrame&w4.BUTTON_RIGHT != ) {
```
aduros, line 123:
```suggestion
    if pressedThisFrame & w4.BUTTON_RIGHT != ) {
```

cli/assets/templates/go/src/main.go
```diff
@@ -1,26 +1,30 @@
...
-    }
+	var gamepad = *w4.GAMEPAD1
+	if gamepad&w4.BUTTON_1 != 0 {
+		*w4.DRAW_COLORS = 4
+	}
 
-    blit(&smiley[0], 76, 76, 8, 8, BLIT_1BPP);
-    text("Press X to blink", 16, 90);
+	w4.Blit(&smiley[0], 76, 76, 8, 8, w4.BLIT_1BPP)
+	w4.Text("Press X to blink", 16, 90)
```
aduros, line 29: 4 spaces indentation plz :nerd_face:

christopher-kleine: No, Go uses Tabs for identation and spaces for aligning variables etc.. Using the "gofmt" tool will enforce this. For that reason, the Spaces vs Tabs war is pretty much non-existent in Go.

cli/lib/png2src.js
```diff
@@ -26,10 +26,14 @@ const %idiomaticName%: [u8; %length%] = [ %bytes% ];
...
-const %name%Flags = %flags%; // %flagsHumanReadable%
-var %name% = [%length%]byte { %bytes% };
+`// %name%
+const (
+    %name%Width = %width%
+    %name%Height = %height%
+    %name%Flags = %flags% // %flagsHumanReadable%
+)
+
+var %name% = [%length%]byte { %bytes% }
```
aduros, line 36: Last nitpick! I think this should remain:

```suggestion
`const %name%Width = %width%
const %name%Height = %height%
const %name%Flags = %flags% // %flagsHumanReadable%
var %name% = [%length%]byte { %bytes% }
```

- More compact output
- Nicer looking output when using multiple images: `w4 png2src assets/*.png --go`. Images are separated by blank lines

cli/assets/templates/go/src/main.go
aduros: Ok, makes sense.

cli/lib/png2src.js
christopher-kleine: Actually I changed it to this *because* this would be more eye-friendly. But I guess that's my personal opinion.

site/docs/guides/user-input.md
christopher-kleine: `gofmt` removes the spaces around the `&` on it's own. No realy use adding them back in. It's the same with the suggestion below.

aduros:
Thanks a lot! :100:

PR #77, opened 2021-09-11
Feature: Source map streaming

eXponenta:
WASM can have a source map.
Fix enable serving a cart.wasm.map by server and patch wasm binary on fly to absolute source map url becuase realtive path produce a `wasm://` protocol that unsupported now.


aduros:
Interesting!

What's the best way to test this feature?

eXponenta:
> Interesting!
> 
> What's the best way to test this feature?

I don't know about other, but AS can generate sourcmap and you can set regular breakpoints to TS source  as in JS, include builtins.

You already merged a template fix for AS taht enable sourmap, but without this PR it cannot working because develope server not response with valid map file :) 

For other languages like C++/RUSat this can Allow DWARF mapping onto source map

https://developer.chrome.com/blog/wasm-debugging-2020/

aduros:
It turns out you can just pass a URL to `--sourceMap` so I ended up doing it that way. Can you please verify that works?

eXponenta:
> It turns out you can just pass a URL to `--sourceMap` so I ended up doing it that way. Can you please verify that works?
> 
> I see the source map working now, but I can't set breakpoints. That might be another issue though.

This will be working for a AS, but can be not for other targets.

And because your dev server is not a clean (it redirect wasm file), need use a fake absolute path.
If you will change a server endpoint you drop a users configulraion, because sourcemap location will again is broken.

And will using a a dynamic port will problem too.

If servers remap a locattion - it should do it in all places that require a absolute path.

My implementation is better anyway.

aduros:
Only AS uses source maps, all the other supported languages use DWARF. If there are new languages that require source maps we can revisit, but for now let's go with the simpler solution with the least moving parts in order to be easier to maintain.

PR #86, opened 2021-09-13
Fix AudioContext

### LINKED ISSUE #84: [bug][safari] wasm4.js crash with iOS < 14.1
### Wasm4 does not work on iOS14.0 or lower

```
Unhandled Promise Rejection: ReferenceError: Can't find variable: AudioContext
```

Tested on simulator with iphone 11Max and iOS14.0 .
It seems that before 14.1 AudioContext was behind a vendor prefix.

source: https://developer.mozilla.org/en-US/docs/Web/API/AudioContext#browser_compatibility

eXponenta:
AudioContext for old IOS is branded

Fix #84

### DIFF
--- a/runtimes/web/src/apu.js
+++ b/runtimes/web/src/apu.js
@@ -22,7 +22,7 @@ const NOISE_LENGTH = 0x8000;
 
 export class APU {
     constructor () {
-        const ctx = new AudioContext();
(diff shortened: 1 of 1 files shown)


FaberVitale:
@eXponenta  👍 

Screenshot 2021-09-13 at 18 50 54

aduros:
That was easy, thanks :tada:

PR #87, opened 2021-09-13
WASM4-81: add Watris CW + CCW spin and switch to D-Pad drop

ubunatic:


aduros:
Looks good to me, thanks! I'll rebuild the cart and copy it to `static/carts/watris.wasm` to update the website.

Issue #82, opened 2021-09-13
Reduce a sence for Virtual Joy


eXponenta:
Problem:
Hard to play when Arrows Up / Left/Right is used in same time because any wrong finger movements produce false input.

I think that corner input should be registered only when  a finger in those zone:
![Propose](https://user-images.githubusercontent.com/12572222/133058360-734facb9-4856-4d4f-9563-c12cabb98ab9.jpg)

What is problem, looks video (focus on joy):
https://youtu.be/dk0RV5x2Bz4?t=16

You can figure out, that when a left/right is pressed , sometimes UP/DOWN is detected too, that produce a wrong movement.
For Snake game this extremely increasing a difficulty.

aduros:
Yeah, I totally agree.

Right now the dpad detection is not great, it just checks distance from the center, and if greater than a dead zone value, assumes the button is pressed. This works well for the action buttons, but makes the dpad diagonals too sensitive.

We should improve it by calculating the dpad angle for the touch (assuming it's within distance of the dpad center).

Thanks for the helpful screenshot by the way :smile:

PR #89, opened 2021-09-13
Reduce virtual DPad sence

eXponenta:
fix for #82


eXponenta:
Ups, has commit for #86

aduros:
Awesome, this is a huge improvement, thanks!

Issue #73, opened 2021-09-09
Regression


eXponenta:
FPS now not synced at 60 FPS.

https://github.com/aduros/wasm4/commit/b6e8b44d69eb39fa4944e979b7a896dde1065150#commitcomment-56199734

For me it always a 30, because elapsed time prcission is 1ms
https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

aduros:
Thanks, I'm taking a look.

Issue #24, opened 2021-08-22
Possibility to set FPS


ttulka:
It would be handy to set FPS from the game code. This would save developers some boilerplate code.

```
export function update(): void {
    // next update will be called in 0.5 sec
    w4.setFPS(30);

    // ...
```

What do you think?

aduros:
What situations would changing the FPS be useful?

ttulka:
Each game has its own update speed, usually depending on the level. For instance, your Tetris speeds up the falling of the bricks. Instead of counting the delay in the game code, one could simply change the FPS like this:

```
w4.setFPS(GAME_FPS[currentLevel]);
```

This is probably not the only one approach, but similar mechanism would be handy for game developers.

aduros:
Hmm, there could maybe be a memory register for it, although for simplicity I feel like it should remain a constant 60 FPS and leave it up to the game to handle. It doesn't seem like that much boilerplate.

Note that tone() durations are measured in "frames" (1/60th of a second), so redefining the length of a frame might create some confusion there.

ttulka:
Maybe is changing FPS not ideal, I agree. Still, I believe *some* out-of-the-box mechanism for controlling the game speed provided by the platform would be really handy.

All in all, it is a problem every game must deal with. Why not solve it once and use it everywhere?

As feedback from other users would be helpful. It'd make sense to postpone this.

leroycep:
I think at very least the speed should be limited. Playing the examples on my main monitors is difficult because it runs at 144 Hz, which makes the games much faster.

FaberVitale:
Documentation currently is currently a bit misleading:
incorrectly assumes that FPS are always at 60 but the amount of times [rAF](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame) runs the callback is tied to several factors:

- is device running in power save mode?
- What's screen refresh rate?
- can the device keep reliable frame rate

For instance on my mac plugging and unplugging the power cord changes the frame rate from 60 to 30.
<img width="560" alt="Screenshot 2021-09-05 at 19 14 24" src="https://user-images.githubusercontent.com/16583125/132135644-38235fdd-1715-411b-842e-10f79d92218c.png">
<img width="562" alt="Screenshot 2021-09-05 at 19 15 05" src="https://user-images.githubusercontent.com/16583125/132135645-2490ffc0-a0e4-4f5a-982f-479237f08d0c.png">

I think there are 2 possible approaches:

[...]

PR #90, opened 2021-09-13
fix: reduce framerate onto 60 FPS

eXponenta:
Realted: #24, #73

Use a implementation from:

https://gist.github.com/addyosmani/5434533#file-limitloop-js-L60

how to check?

Run chrome without v-sync:
https://gist.github.com/brunosimon/c15e7451a802fa8e34c0678620022f7d

Open any game.

Video:  https://youtu.be/VxCclCHjwII
LEFT- fixed, RIGHT - current


aduros:
Looks good, thanks!

PR #91, opened 2021-09-14
Added option "--open" to open browsers

christopher-kleine:
The commands "watch" and "run" open the browser (or a new tab) by default.
I often have this tab already open. Executing those commands open *another* tab.

That's why I've added an option to open the browser if desired:

`w4 watch --open` and `w4 run --open CART`

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-const { program } = require("commander");
+const { program, Option } = require("commander");
 const pkg = require('./package.json');
 const { supportedIconExtensions } = require('./lib/utils/icon');
 
@@ -17,7 +17,11 @@ program.command("new <directory>")
 
 program.command("watch")
     .description("Rebuild and refresh when source code changes")
-    .option("--open", "Opens the browser")
+    .addOption(
+        new Option("-n, --no-open", "Doesn't open the browser")
+        .env("W4_NO_OPEN")
+        .default(false)
+    )
     .action(opts => {
         const watch = require("./lib/watch");
         watch.start(opts);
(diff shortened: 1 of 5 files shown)


aduros:
What do you think about opening by default and adding a `--no-open` to disable it? Opening is a little friendlier to new users.

FaberVitale:
@aduros  
commander adds -no-<bool> option automatically: https://www.npmjs.com/package/commander#common-option-types-boolean-and-value

You could set the default value to `true`

christopher-kleine:
Typing `--no-open` is more work than just closing the browser/tab.
That's actually why I went to made it `--open` instead.
That's still faster than to to open your browser and type in the URL.

If that's still too much, how about adding a shorthand for `--open`? Like `-b` for Browser?

aduros:
We could add `-n` as a shorthand to `--no-open`.

By the way, does the live-reloading work for you? The idea is not that you need to re-run `w4 run` after every recompile, but only once. Assuming it's working correctly it should watch for file changes to the cart and reload the browser automatically.

christopher-kleine:
> We could add -n as a shorthand to --no-open.

Deal. And done. But since I _realy_ want "no-open" to be the default on my PC, I made it possible to use an env-variable to override it.

```bash
w4 run build/cart.wasm # Opens the browser
w4 run -n build/cart.wasm # Does NOT opens the browser
w4 run --no-open build/cart.wasm # Does NOT opens the browser
W4_NO_OPEN w4 run build/cart.wasm # Does NOT opens the browser
```

aduros:
> But since I realy want "no-open" to be the default on my PC, I made it possible to use an env-variable to override it.

Ooh, that's smart :smiley: 

Thanks!

PR #92, opened 2021-09-14
refactor(png2src): change rust template dimensions from u32 to usize

FaberVitale:
Description

`usize` is used to index vectors, arrays, strings and slices.
It is system dependent either 32bit or 64bit.

References:

* https://doc.rust-lang.org/std/primitive.usize.html

### DIFF
--- a/cli/lib/png2src.js
+++ b/cli/lib/png2src.js
@@ -19,9 +19,9 @@ const uint8_t %name%[%length%] = { %bytes% };
 `,
 
     rust: 
-`const %idiomaticName%_WIDTH: u32 = %width%;
-const %idiomaticName%_HEIGHT: u32 = %height%;
-const %idiomaticName%_FLAGS: u32 = %flags%; // %flagsHumanReadable%
+`const %idiomaticName%_WIDTH: usize = %width%;
+const %idiomaticName%_HEIGHT: usize = %height%;
+const %idiomaticName%_FLAGS: usize = %flags%; // %flagsHumanReadable%
 const %idiomaticName%: [u8; %length%] = [ %bytes% ];
 `,  
 


aduros:
It looks like this causes a compiler error when using the parameters with `blit()`, which expects `u32` for width/height/flags.

What are the benefits of using `usize` instead of `u32`?

eXponenta:
usize used as pointer because represent location in memory.
You can use a u32 as pointer, but pointer will less that u32 limit on wasm64 (that support 64 bit pointer) and casted onto usize implicit.

No make sense which use a u32 or usize, because anyway all will be casted onto u32 or u16 (i not see any reason use a u32 for width/height/flags) explicit.

BUT! 
structure size will depend what is types used, and for u32 you alloc a 4+4+4 + (width * height >> 2) bytes, but when used a usize - 8 + 8 + 8  + (width * height >> 2)
12 bytes overhead for nothing.

I propose use a u16 or u8 for width/height/flags with implicit cast where it used.

because really there are not reason use a frames greater than 256pixels, and flags store 1 bit of information.

If you want pack metadata onto image data buffer, this will do nothing because metadata will store same values with same dimensions as in struct because should be aligned :)

FaberVitale:
> It looks like this causes a compiler error when using the parameters with blit(), which expects u32 for width/height/flags.

There's no implicit numerical promotion in rust. 
To change numeric type you have to **explicitly** convert it using either `as` operator or using traits  like `From` or `Into`.

> What are the benefits of using usize instead of `u32?

It gets rather annoying to manually have convert from `u32` to `usize` when you need to use that number as index or bound of a slice or when you need are doing math ops that have different types.

It would be better to convert `SCREEN_SIZE` and those params used by `blit` and `blit_sub` to `usize`.

eXponenta:
You always can use a custom template =)
And custom wasm4 bindings.

In first - you try to use a usize in enviroument where a wee_alock not working as expected, and second - this is Rust, you by design should to Cry when use it!😅

FaberVitale:
> You always can use a custom template =)
> And custom wasm4 bindings.
> 
> In first - you try to use a usize in enviroument where a wee_alock not working as expected, and second - this is Rust, you by design should to Cry when use it!😅

I do not understand this comment:
it is not the first PR that suggests changes to the template provided.

The goal of these PRs is to improve the usability of the template provided.
If the maintainer deems it valid it will be merged, otherwise it won't.

Nobody is "crying" over anything.

aduros:
> It would be better to convert SCREEN_SIZE and those params used by blit and blit_sub to usize.

Ok, if you feel this would be an improvement, let's do it. Can you update the PR? The only thing I care about is avoiding annoying casts.

> I propose use a u16 or u8 for width/height/flags with implicit cast where it used, because really there are not reason use a frames greater than 256pixels, and flags store 1 bit of information.

WebAssembly parameters types are only 32 bit or 64 bit. u16/u8 exists in AssemblyScript, but these are equivalent to u32 when compiled. As far as memory usage goes, these constants are inlined at the call site, so their type shouldn't make any actual difference in the compiled output.

eXponenta:
> WebAssembly parameters types are only 32 bit or 64 bit. u16/u8 exists in AssemblyScript, but these are equivalent to u32 when compiled. As far as memory usage goes, these constants are inlined at the call site, so their type shouldn't make any actual difference in the compiled output.

Yep, but who use a constant call in game? This is not fully useful, because for each new sprite you should create new function that wrap a `blit` function and store it in function pointer.
I think anyway when you have more that 2-3 sprites you will use a structure which store this value and 
Wasm cannot use pointer onto binary and you anyway should allocate a memory block for structure.

AssemblyScript, Rust, Go, C - no make sense.

FaberVitale:
I've changed my my on this.

`usize` becomes `wasm - i32` but it is probably better for readability reasons to use `u32` instead of `usize`.

PR #94, opened 2021-09-16
Setting default language using env vars

christopher-kleine:
Setting the env var "W4_LANG" changes the default value for "new" and "png2src".
For this, a new option was added: "--lang".
This option can only have these options:

- as
- assemblyscript
- c
- rs
- rust
- go

In case no value was provided by either env or cli,
it falls back to "as".
It can still be overriden by using the options --as etc.

Examples:
```bash
w4 new snake # Creates "snake" using AssemblyScript
w4 new --go snake # Creates "snake" using Go
W4_LANG=go w4 new snake # Creates "snake" using Go
W4_LANG=go w4 new --rust snake # Creates snake using Rust
```

The same is true for `png2src`.

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -6,10 +6,16 @@ const { supportedIconExtensions } = require('./lib/utils/icon');
 
 program.command("new <directory>")
     .description("Create a new blank project")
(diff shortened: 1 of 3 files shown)


aduros:
Thanks!

PR #95, opened 2021-09-17
Added "w4 init" - Fixes #50

### LINKED ISSUE #50: `w4 init` command
Similar to `w4 new` but instead initialize the project in the current directory. Shouldn't be hard to implement

christopher-kleine:
In essence a "w4 init" is an alias to "w4 new ."

To achieve this and reduce code duplication, I've created a new
function called "blankProject" inside cli.js.
It takes a program.command and attaches the options of the "new" command
This function then is called for "new" and "init".

To make it a bit simpler still, I've also added a "real" alias to
"new" called "create".

Further "changes"to the source code are due to auto-formatting it.

// Edit: Yay, another low-hanging fruit solved!

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -4,29 +4,45 @@ const { program, Option } = require("commander");
 const pkg = require('./package.json');
 const { supportedIconExtensions } = require('./lib/utils/icon');
 
-program.command("new <directory>")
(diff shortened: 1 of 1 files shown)


aduros:
Awesome, thanks for taking care of this one! :tada:

PR #104, opened 2021-09-20
feat: Gamepad

eXponenta:
Fix #104 

Support native gamepad.
Now only first gamepad is used.

DPAD + LEFT AXIS - movement
X, O (DS4 mapping) - X, Z
LEFT/RIGHT Triggers - X, Z

Sometimes joys swap a buttons order for right buttons panel, need investigate this case.

This is feature a runtime, this means that need rebundle games.

UPD:
If you trigger a keyboard or gamepad in game - last input device will be used. You can change gamepad/keyboard on fly.
First connected gamepad is used on multiple gamepad system , and replaced on latest available after disconnection.

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -274,15 +274,6 @@ async function loadCartWasm () {
     }
 
     window.addEventListener("gamepadconnected", (e) => {
-        const gp = navigator.getGamepads()[e.gamepad.index];
-        console.log(
-          "Gamepad connected at index %d: %s. %d buttons, %d axes.",
-          gp.index,
-          gp.id,
-          gp.buttons.length,
-          gp.axes.length
-        );
-
         gamepads ++;
     });
 
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -288,17 +288,7 @@ async function loadCartWasm () {
 
     window.addEventListener('gamepaddisconnected', (e) => {
         gamepads --;
-
-        // https://www.w3.org/TR/gamepad/#remapping
-        // DPAD + AXIS
-        runtime.maskGamepad(0, constants.BUTTON_UP, false);
-        runtime.maskGamepad(0, constants.BUTTON_DOWN, false);
-        runtime.maskGamepad(0, constants.BUTTON_LEFT, false);
-        runtime.maskGamepad(0, constants.BUTTON_RIGHT, false);
-
-        // X, O + Triggers
-        runtime.maskGamepad(0, constants.BUTTON_X, false);
-        runtime.maskGamepad(0, constants.BUTTON_Z, false);
+        runtime.maskGamepad(0,0, false);
     });
 
     const dpad = document.getElementById("gamepad-dpad");
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -197,8 +197,17 @@ async function loadCartWasm () {
         122: requestFullscreen, // F11
     };
 
+    let isKeyboard = true;
+
     const onKeyboardEvent = event => {
         const down = (event.type == "keydown");
+
+        if (!isKeyboard) {
+            // reset joy pad state
+            runtime.setGamepad(0,0);
+        }
+
+        isKeyboard = true;
 
         // Poke WebAudio
         runtime.unlockAudio();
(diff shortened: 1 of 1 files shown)


runtimes/web/src/index.js
```diff
@@ -245,6 +245,62 @@ async function loadCartWasm () {
...
+        // https://www.w3.org/TR/gamepad/#remapping
+        // DPAD + AXIS
+        runtime.maskGamepad(0, constants.BUTTON_UP, false);
+        runtime.maskGamepad(0, constants.BUTTON_DOWN, false);
+        runtime.maskGamepad(0, constants.BUTTON_LEFT, false);
+        runtime.maskGamepad(0, constants.BUTTON_RIGHT, false);
+
+        // X, O + Triggers
+        runtime.maskGamepad(0, constants.BUTTON_X, false);
+        runtime.maskGamepad(0, constants.BUTTON_Z, false);
```
aduros, line 301:
```suggestion
        runtime.setGamepad(0, 0);
```

runtimes/web/src/index.js
```diff
@@ -278,7 +334,6 @@ async function loadCartWasm () {
             const DPAD_MAX_DISTANCE = 100;
             const DPAD_DEAD_ZONE = 10;
             const BUTTON_MAX_DISTANCE = 50;
-            const DPAD_ACTIVE_ZONE = 3 / 5; // cos of active angle, greater that cos 60 (1/2)
```
aduros, line 281: Did you mean to revert these virtual gamepad changes?

eXponenta: ahhh. I synced wrong version (i had a dPAD in isolated branch, but commit onto main), wait.

runtimes/web/src/index.js
```diff
@@ -245,6 +245,62 @@ async function loadCartWasm () {
...
+        // DPAD + AXIS
+        runtime.maskGamepad(0, constants.BUTTON_UP, buttons[12].pressed || axes[1] < -0.5);
+        runtime.maskGamepad(0, constants.BUTTON_DOWN, buttons[13].pressed || axes[1] > 0.5);
+        runtime.maskGamepad(0, constants.BUTTON_LEFT, buttons[14].pressed || axes[0] < -0.5);
+        runtime.maskGamepad(0, constants.BUTTON_RIGHT, buttons[15].pressed || axes[0] > 0.5);
+
+        // X, O + Triggers
+        // NOTE: for XBox360 a triggers is 6 and 7
+        runtime.maskGamepad(0, constants.BUTTON_X, buttons[0].pressed || buttons[6].pressed);
+        runtime.maskGamepad(0, constants.BUTTON_Z, buttons[1].pressed || buttons[7].pressed);
```
aduros, line 273: Instead of making lots of calls to maskGamepad, let's start with `let gamepad = 0`, assign each bit (eg: `if (buttons[0].pressed) gamepad |= constants.BUTTON_X`) and then apply it with `runtime.setGamepad(0, gamepad)`.

runtimes/web/src/index.js
```diff
@@ -245,6 +245,62 @@ async function loadCartWasm () {
...
+
+    window.addEventListener("gamepadconnected", (e) => {
+        const gp = navigator.getGamepads()[e.gamepad.index];
+        console.log(
+          "Gamepad connected at index %d: %s. %d buttons, %d axes.",
+          gp.index,
+          gp.id,
+          gp.buttons.length,
+          gp.axes.length
+        );
```
aduros, line 284: Please comment out this logging

runtimes/web/src/index.js
eXponenta: This will not working if gamepad buttons will used together with keyboard.
But yep, not make sence, because this is explict case and should be fixed if a multiple gamepads is mapped onto gamepad/keyboard .

runtimes/web/src/index.js
```diff
@@ -245,6 +245,62 @@ async function loadCartWasm () {
...
     window.addEventListener("keyup", onKeyboardEvent);
 
+    let gamepads = 0;
+    function processGamepad() {
+        if (gamepads === 0) {
+            return;
+        }
+
+        const gamepad = navigator.getGamepads()[0];
+        if (!gamepad) {
```
aduros, line 255: I think there's a bug here where when the gamepad is disconnected and reconnected, it'll stop working because gamepad index 0 will still be null.

Could we store the last gamepad index connected in the `gamepadconnected` handler, and then use it in `processGamepad`?

eXponenta: This is problematic case. I have a 2 gamepads and, 1 is real, 2 is virtual.
And a lot of gamepad mapping programms replace a 1 gamepad.
If we will use a latest connected gamepad - it will not working.

eXponenta: And looks like that when gamepad is disconnected it changings a gamepad list and it never can be a gaps.

aduros: We could make it the oldest still connected gamepad index, instead of the most recently connected. It should just be a valid index.

runtimes/web/src/index.js
aduros: > This will not working if gamepad buttons will used together with keyboard.

I think that's the case with `maskGamepad` as well since it will clear the button bit if pressed is false.

runtimes/web/src/index.js
eXponenta: Yep, i do this.
I checked a bechaviour, i have a 2 gamepads (BL and usb) and there are a difference.
XBox gamepad is always is 0, but USB is not change a index when XBox is unplugging.

eXponenta: Note.
maskGamepad working some strange. Because it use a ` buttons &= ~maks` for rest need use a `maskGamepad(0, ~1, false)`  instead of `maskGamepad(0,0)`

eXponenta:
More compicated changes.
Now if used a gamepad but you trigger a keyboad - keyboard will used and vice versa.

First connected gamepad is used on multiple gamepad system , and replaced on latest available after disconnection.

runtimes/web/src/index.js
```diff
@@ -245,6 +254,83 @@ async function loadCartWasm () {
...
+        const gamepad = navigator.getGamepads()[usedGamepad];
+        if (!gamepad) {
+            return;
+        }
+
+        const buttons = gamepad.buttons;
+        const axes = gamepad.axes;
+        
+        // not all gamepads map DPAD to 12,13, 14, 15 buttos
+        const dpad = buttons.length > 12;
```
aduros, line 273: Would it be better to instead return early if `gamepad.mapping != "standard"`?

runtimes/web/src/index.js
```diff
@@ -245,6 +254,83 @@ async function loadCartWasm () {
...
+
+
+        let buttonMask = 0;
+        buttonMask |= constants.BUTTON_UP * up;
+        buttonMask |= constants.BUTTON_DOWN * down;
+        buttonMask |= constants.BUTTON_LEFT * left;
+        buttonMask |= constants.BUTTON_RIGHT * right;
+
+        buttonMask |= constants.BUTTON_X * x;
+        buttonMask |= constants.BUTTON_Z * z;
```
MaxGraey, line 295: what about this?

```suggestion
        const up = dpad && buttons[12].pressed || axes[1] < -0.5;
        const down = dpad && buttons[13].pressed || axes[1] > 0.5;
        const left = dpad && buttons[14].pressed || axes[0] < -0.5;
        const right = dpad && buttons[15].pressed || axes[0] > 0.5;

        // X, O + Triggers
        // NOTE: for XBox360 a triggers is 6 and 7
        const x = buttons[0].pressed || buttons[6].pressed;
        const z = buttons[1].pressed || buttons[7].pressed;

        let buttonMask = 0;
        buttonMask |= constants.BUTTON_UP & -up;
        buttonMask |= constants.BUTTON_DOWN & -down;
        buttonMask |= constants.BUTTON_LEFT & -left;
        buttonMask |= constants.BUTTON_RIGHT & -right;

        buttonMask |= constants.BUTTON_X & -x;
        buttonMask |= constants.BUTTON_Z & -z;
```

runtimes/web/src/index.js
eXponenta: This is unsafe.
mapping can be non-standart but a buttons exist and back.

Then you suggest prevent support a gamepads like this:)

image

aduros: If the mapping is non-standard, how do we know that `gamepad.buttons[14]` means "left"? The handling seems to already rely on the standard mapping.

But anyways, this is minor, we can fix it up later if necessary.

PR #111, opened 2021-09-26
fix(rust): allocation not working

### LINKED ISSUE #36: Rust `text` or `trace` and `format!` throws `RuntimeError: unreachable executed`
When attempting to draw dynamic text to the screen in rust, the runtime aborts entirely.

```rust
// lib.rs
mod wasm4;
use wasm4::*;

#[no_mangle]
fn update() {
    unsafe {
        *DRAW_COLORS = 2;
    }
    text(format!("{:02x}", unsafe {CT}).as_str(), 10, 10);
    unsafe { CT += CT.wrapping_add(1) }
}
```

Error in the console:
```
Uncaught (in promise) RuntimeError: unreachable executed
    update http://localhost:4444/wasm4.js:1
    t http://localhost:4444/wasm4.js:1
    <anonymous> http://localhost:4444/wasm4.js:1
    async* http://localhost:4444/wasm4.js:1
    <anonymous> http://localhost:4444/wasm4.js:1
wasm4.js line 1 > WebAssembly.instantiate:15530:1
    <anonymous> http://localhost:4444/wasm4.js:1
    AsyncFunctionNext self-hosted:692
    (Async: async)
    <anonymous> http://localhost:4444/wasm4.js:1
    <anonymous> http://localhost:4444/wasm4.js:1
```

### LINKED ISSUE #37: [Rust] cannot allocate a String in update

### LINKED ISSUE #42: [Rust] Handling of panics by passing them to trace()
Adding this to `start()` will hook all panic messages (probably), unless it panics during a panic in which case it just crashes. If memory problems are present it will crash on alloc so we can't rely on `.to_string()`.

```Rust
#[no_mangle]
fn start() {
    #[cfg(debug_assertions)]
    panic::set_hook(Box::new(|info| {
        if let Some(s) = info.payload().downcast_ref::<&str>() {
            trace(s);
        } else {
            trace("A panic occurred somewhere...")
        }

        /* // I am unsure of what wizardry wasmbindgen uses but backtraces don't seem to work independently.
        trace("Stack:\n");
        let backtrace = Backtrace::force_capture();
        match backtrace.status() {
            std::backtrace::BacktraceStatus::Unsupported => trace("No backtrace support"),
            std::backtrace::BacktraceStatus::Disabled => {
                trace("Something disabled the backtracing")
            }
            std::backtrace::BacktraceStatus::Captured => {
                for frame in backtrace.frames() {
                    trace(&format!("{:?}", frame));
                }
            }
            _ => trace("The compiler demanded I add this. If you see this assume everything has gone wrong."),
        };
        */

        //trace(&info.to_string())
        // .to_string() will just crash if allocations are not working
    }));
}
```
An `error(&str)` function could be added so that stack logs could be passed to the console as errors.

FaberVitale:
### Description

Adds [buddy-alloc](https://crates.io/crates/buddy-alloc) as optional default allocator and increases the stack size.

### Issues

- Closes #42 
- Closes #37 
- Closes #36

### Related PR

- #78

### See also

- https://github.com/rustwasm/wasm-bindgen/issues/1389#issuecomment-476224477
- https://github.com/rustwasm/wee_alloc/issues/61
- https://github.com/rustwasm/wee_alloc/issues/88#issuecomment-581404618


cli/assets/templates/rust/src/alloc.rs
```diff
@@ -0,0 +1,29 @@
...
+const FAST_HEAP_SIZE: usize = 4 * 1024; // 4 KB
+const HEAP_SIZE: usize = 16 * 1024; // 16 KB
+const LEAF_SIZE: usize = 16;
+
+static mut FAST_HEAP: [u8; FAST_HEAP_SIZE] = [0u8; FAST_HEAP_SIZE];
+
+#[global_allocator]
+static ALLOC: NonThreadsafeAlloc = unsafe {
+    let fast_param = FastAllocParam::new(FAST_HEAP.as_ptr(), FAST_HEAP_SIZE);
+    let buddy_param = BuddyAllocParam::new(heap_ptr!(u8), HEAP_SIZE, LEAF_SIZE);
```
aduros, line 27: Nice!!

Can we use `__heap_base` to dynamically size the heap size instead of needing a `HEAP_SIZE` constant? Since the total memory in WASM-4 is 64 KB, we should be able to calculate the heap size to be `65535 - __heap_base`?

cli/assets/templates/rust/src/lib.rs
```diff
@@ -1,19 +1,14 @@
...
 const SMILEY: [u8; 8] = [
-    0b11000011,
-    0b10000001,
-    0b00100100,
-    0b00100100,
-    0b00000000,
-    0b00100100,
-    0b10011001,
-    0b11000011,
+    0b11000011, 0b10000001, 0b00100100, 0b00100100, 0b00000000, 0b00100100, 0b10011001, 0b11000011,
```
aduros, line 7: Let's keep the formatting for this as it was, I think it's important for beginners to be able to somewhat visualize the sprite.

cli/assets/templates/rust/src/wasm4.rs
```diff
@@ -123,8 +123,9 @@ extern "C" {
 }
 
 /// Draws text using the built-in system font.
-pub fn text(text: &str, x: i32, y: i32) {
-    unsafe { extern_text(text.as_ptr(), text.len(), x, y) }
+pub fn text<T: AsRef<str>>(text: T, x: i32, y: i32) {
+    let text_ref = text.as_ref();
+    unsafe { extern_text(text_ref.as_ptr(), text_ref.len(), x, y) }
```
aduros, line 128: Interesting, does this allow auto-casting from other string types?

cli/assets/templates/rust/.cargo/config.toml
```diff
@@ -10,8 +10,8 @@ rustflags = [
     "-C", "link-arg=--initial-memory=65536",
     "-C", "link-arg=--max-memory=65536",
 
-    # Reserve 1024 bytes of stack space, offset from 6580
-    "-C", "link-arg=-zstack-size=7584",
+    # Reserve 1228 bytes of stack space, offset from 6580
+    "-C", "link-arg=-zstack-size=7808",
```
aduros, line 14: Should we bump this to 2048+ bytes just to be safe?

cli/assets/templates/rust/src/lib.rs
FaberVitale: I'm not so sure about this:

that's the output of [rustfmt](https://github.com/rust-lang/rustfmt):
conforming to language standards, facilitates communication between developers.

Moreover if you develop using an IDE you'll probably get this piece of code formatted properly as soon as you save or commit.
Have you considered using comments to prevent this behavior?

cli/assets/templates/rust/src/wasm4.rs
FaberVitale: `AsRef<str>` is a [common pattern in rust api design](https://www.philipdaniels.com/blog/2019/rust-api-design/).
The gist is that you can **also** pass `String` types directly without explicitly passing a reference.
It reduces the clutter especially when using the macro `format!`

### Before

```rust
let formatted = format!("{:?}",  SMILEY);

text("string slice", 50, 50);
text(&formatted, 20, 20);
``` 


### After

```rust
text("string slice", 50, 50);
text(format!("{:?}",  SMILEY), 20, 20);
```

I've also evaluated also the possibility to add `text!` and `trace!` macros, simple wrappers for `text(format!([...]))`,  but I'm not convinced that we should provided that as default because it is not `free` in terms of bytes used.

cli/assets/templates/rust/.cargo/config.toml
FaberVitale: I do not think that there's a perfect number or at least I do not know a 
I think that we should just add a comment that suggests to bump this value in case the framebuffer gets corrupted.

cli/assets/templates/rust/src/alloc.rs
FaberVitale: I've tried to achieve that using several strategies but even in nightly mode I get an error when I try it.

### Explanation

`__heap_base`  taken from the[`extern` block is **always 0**](https://github.com/rustwasm/wee_alloc/issues/88#issuecomment-581547608) but you could extract its value using a simple cast `as u32` unfortunately
 I've not found a way to write `65535 - __heap_base`.

#### Things I've tried

1.  `BuddyAllocParam::new(heap_ptr!(u8), 65535 - (heap_ptr!(u8) as usize), LEAF_SIZE);`

```
Compiler error:
pointers cannot be cast to integers during const eval.
at compile-time, pointers do not have an integer value
avoiding this restriction via `transmute`, `union`, or raw pointers leads to compile-time undefined behavior
```

2.  `BuddyAllocParam::new([...](heap_ptr!(u8).offset_from(0 as *const u8) as usize), [...])`

Requires switch to nightly and flag `#![feature(const_ptr_offset_from)]`

```
Compiler error:
null pointer is not a valid pointer for this operation
more info here: rustc --explain E0080
```

Essentially rust does not support pointer arithmetic when working with `const` values.

I'm wondering if it would be safer to use two array instead, like the snippet you showed in the other thread.
What do you think?

aduros: I think the idea about using `__heap_base` was that so we could dynamically size the heap. If that doesn't work, I guess it might be better to just use a statically allocated array.

Here's a thought... do we even need to calculate a valid heap size? What happens if we just set the `HEAP_SIZE` constant to something huge like a megabyte? It will of course actually run out of memory far sooner, but the wasm runtime can catch that, rather than buddy-alloc.

cli/assets/templates/rust/src/lib.rs
aduros: Would `#[rustfmt::skip]` work here?

cli/assets/templates/rust/src/wasm4.rs
aduros: Good to know, thanks!

cli/assets/templates/rust/src/alloc.rs
FaberVitale: > Here's a thought... do we even need to calculate a valid heap size? What happens if we just set the HEAP_SIZE constant to something huge like a megabyte?

I've tried with something far smaller, `65535`, but it does not work.

### Test code

```rust
    let mut vecc: Vec<u32> = vec![1, 2, 3];

    for n in 0..20 {
        trace(format!("{:?}", n));
        vecc.push(n);
    }
```

it crashes as soon you reach the 8th push:
I haven't checked the source code but I assume that it tries to ensure capacity and fails.

Using a static heap instead works and the vector grows much further.
Moreover it is the way officially supported by the [allocator](https://github.com/jjyr/buddy-alloc/blob/master/examples/non_threadsafe_demo.rs).

aduros:
Bummer that `__heap_base` dynamic sizing didn't work out, but I'm happy with how it turned out.

Thanks so much for your work on this! It's great to finally have dynamic memory allocation all sorted out across all 4 supported wasm languages. :smile:

PR #113, opened 2021-10-05
Added Wormhole game

thedavesims:


thedavesims:
There is an issue with Wormhole running when I checked the preview, I'll update w4 and build it again. It looks like the circle drawing code has changed between by build.

aduros:
Oh yeah, `oval()` now supports a fill color in draw colors slot 1, so I think you'll need to make sure that slot is zero to prevent it from filling. Eg:

```c
*DRAW_COLORS = 0x20;
```

PR #118, opened 2021-10-15
Added bound() function and bounds checks

thedavesims:
Added a bounds() function to make bounds checking more readable in the code and applied it in w4_framebufferOval.
I'm not sure if this is the approach you want to take with it.


thedavesims:
Yes, this fixed it.

PR #119, opened 2021-10-15
Added conditionals for macOS

thedavesims:
Added conditionals to enable using OpenGL 2 on macOS as it does not support OpenGL ES 2.0.
Just showing the changes I made to get it to run on macOS.


aduros:
Thanks, I committed something that should help and be more cross-platform: https://github.com/aduros/wasm4/commit/f67c123010acc0f72a5894f254bac0223076430d

Can you please let me know if it works on mac?

PR #122, opened 2021-10-16
Added type casts

thedavesims:
Added type casts to remove compiler warnings.


thedavesims:
Also added a fix for the palette not resetting after a hotswap when using 'w4 watch'.

PR #134, opened 2021-10-30
Add Lake Shooter cartridge to site

jimmycartrette:


aduros:
Challenging and a lot of fun! It seems very faithful to the classic Atari version :space_invader: 

Would it be possible to restart the game at the game over screen without needing to press R to hard reset? (which only works on desktop web)

Also minor but "press to play" could be changed to "press X to play" so that players know how to start the game :smiley:

jimmycartrette:
Thanks! Submitting some updates for that now

PR #145, opened 2021-11-10
add wasm4 port of 256 byte TIC80 game Skip Ahead

exoticorn:


eXponenta:
I can't restart game after loose.
I not see any score.

exoticorn:
You can now restart with Z and there is a score display. And I managed to keep it all in 512 bytes \o/

aduros:
Absolutely incredible feat of a game in **512 bytes**!! :tada:

PR #153, opened 2021-11-13
Tweaks for D

aduros:
@tjpalmer Can you please review these small changes when you get a chance?


tjpalmer:
I'm still an expert in neither D nor WASM-4, but they look good. I also just put in my own PR for this branch if you want to include some extra tweaks along with these. It fixes an audio example mistake that I missed earlier and also makes some indentation more consistent.

There are other files in the code base with mixed tabs and spaces, but I limited my scope to things already going on here.

PR #162, opened 2021-11-18
docs: rust tutorial

FaberVitale:
### Description

Let's do it!

----

### References:

- https://www.google.com/search?client=firefox-b-d&q=rust+install
- https://doc.rust-lang.org/book/
- https://github.com/christopher-kleine/wasm-4-tutorial-games/tree/main/snake-rust

### DIFF
--- a/site/docs/tutorials/snake/collision-detection.md
+++ b/site/docs/tutorials/snake/collision-detection.md
@@ -34,7 +34,7 @@ fruit.X = rnd(20)
 fruit.Y = rnd(20)
 ```
 
-In it's final form, it could look like this:
+In its final form, it could look like this:
 
 ```typescript {4-9}
     if (frameCount % 15 == 0) {
@@ -88,7 +88,7 @@ fruit.X = rnd(20)
 fruit.Y = rnd(20)
 ```
 
-In it's final form, it could look like this:
+In its final form, it could look like this:
 
 ```go {4-8}
 	if frameCount%15 == 0 {
@@ -135,7 +135,7 @@ Just this however will realocate the fruit in the same places for every run beca
 math.randomseed(frame_count)
 ```
 
-In it's final form, it could look like this:
+In its final form, it could look like this:
 
 ```lua {4-9}
   if frame_count % 15 == 0 then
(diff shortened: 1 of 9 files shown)


FaberVitale:
I've completed the rust tutorial and I've added a note that explains how to [install binaryen](https://github.com/aduros/wasm4/pull/162/files#diff-77d8ef51e717bf3a232a56e3f8a28c7ad89a2ff3ed38fad89e9e9616c9ced2a7R212) and how to  use it in [publish-your-game.md](https://github.com/aduros/wasm4/pull/162/files#diff-7f87234724bb9c99617db902700dda34ee90b50ae0e8ae0db74da85bf1dfdf3bR81).

jamesmcm:
This is awesome. It might be worth mentioning wasm-snip too (see issue #238 )  since it can get you a smaller binary (796 bytes vs 1.8kb on the basic tutorial for example) compared to wasm-opt (both much smaller than the 27k without optimisation though!) - at the cost of stripping out all of the std formatting and panicking support (maybe there is a way of doing this with wasm-opt too though?).

FaberVitale:
> This is awesome. It might be worth mentioning wasm-snip too (see issue #238 ) since it can get you a smaller binary (796 bytes vs 1.8kb on the basic tutorial for example) compared to wasm-opt (both much smaller than the 27k without optimisation though!) - at the cost of stripping out all of the std formatting and panicking support (maybe there is a way of doing this with wasm-opt too though?).

> at the cost of stripping out all of the std

I have not tried `wasm-snip` but it can be added as a mention inside `site/docs/tutorials/snake/publish-your-game.md`.

OliveIsAWord (changes requested):
One obvious change is replacing this line...

```rust
self.frame_count = self.frame_count.overflowing_add(1).0;
```

with this...

```rust
self.frame_count = self.frame_count.wrapping_add(1);
```

It still seems rather verbose, and I wonder if it would be more helpful to just use a regular increment. As is, it would take years for the frame_counter to overflow, and with a u64, it would take billions of years. I prefer explicitly wrapping, but I worry it might not be as helpful for beginners.

Also, there are many erroneous uses of `it's` rather than `its` throughout the tutorial. These should be fixed.

aduros:
Amazing work, thank you Faber!

PR #166, opened 2021-11-21
feat: add `wasm4::seed` in rust template

FaberVitale:
### Description

Adds binding to `seed` in rust template.

### Why

Seed is a randomness source that can be fed to a pseudo random number generator.

### Usage with a prng

Required packages: 

- [rand](https://crates.io/crates/rand) 
- [rand_pcg](https://crates.io/crates/rand_pcg)

```rust
mod wasm4;

use rand::prelude::*;
use rand_pcg::Pcg64;

#[no_mangle]
fn update() {
    let rng = Pcg64::seed_from_u64(wasm4::seed());

    rng.gen_range(0..5); // returns a number between  0 included and 5 excluded

    rng.gen_range(0..=5); // returns a number between  0 included and 5 included
}

```

### Simple usage demo

Nov-21-2021 10-49-52

### Prng usage demo
rand-Nov-21-2021 11-29-27

### DIFF
--- a/cli/assets/templates/rust/src/wasm4.rs
+++ b/cli/assets/templates/rust/src/wasm4.rs
@@ -217,3 +217,13 @@ extern "C" {
     #[link_name = "traceUtf8"]
     fn extern_trace(trace: *const u8, length: usize);
 }
+
+/// Returns a random `u64`.
+pub fn seed() -> u64 {
+    unsafe { extern_seed() as u64 }
+}
+
+extern "C" {
+    #[link_name = "seed"]
+    fn extern_seed() -> f64;
+}


aduros:
`seed` was a quick hack that was added for AssemblyScript that I kind of regret. I'd like to add netplay support in the future, and that means games need to be fully deterministic. Also from a "fantasy console aesthetic" perspective, I don't like the idea of a function that enables true randomness. One way that retro games handle randomness is by counting the number of frames since the first button press as a seed, for example. There's a real charm in old games that are susceptible to RNG manipulation by speed runners.

The fact that `seed` exposes the system time is also something that can be misused (and breaks netplay even more).

I'm strongly considering removing `seed` and maybe for convenience adding a new memory register called `FRAME_COUNT` or something that can be used to easily seed an RNG.

aduros:
@claudiomattera I'd like to remove `seed` but it looks like it's being used by minesweeper. Would it be possible to rewrite to use a seed based on frames elapsed, the mouse position, or even just a constant?

claudiomattera:
> @claudiomattera I'd like to remove `seed` but it looks like it's being used by minesweeper. Would it be possible to rewrite to use a seed based on frames elapsed, the mouse position, or even just a constant?

Yes.
I will make a new pull request with a new cartridge in the weekend.

Issue #147, opened 2021-11-11
Nim support


aduros:
https://github.com/Yardanico/wasm4nim/

- [x] `w4 new`: Add new project template
- [x] `w4 watch`: Add new filename suffix + build/run support
- [x] `w4 png2src`: Add new language parameter
- [x] Website: Setup documentation
- [ ] Website: Translated code snippets for all guide pages

Yardanico:
Hi! I'd be glad to add Nim to wasm4. Some questions:
- Is it better to make code a bit more idiomatic than having a 1:1 mapping with the WASM-4 itself? For example, defines like `BUTTON_1` or `BUTTON_LEFT` can be made into a Nim enum, but then the names (and usage of functions) will differ. On the other hand, having almost the same interface for all languages can make writing examples much easier, so I'm not sure.

- I assume that wasm4 is a monorepo and so I'll have to commit all changes to the wrapper (although it's so simple that I think it won't have a lot of frequent changes) via the PRs? This is not an issue for me, so I'm just asking out of curiosity :)

aduros:
>     * Is it better to make code a bit more idiomatic than having a 1:1 mapping with the WASM-4 itself? For example, defines like `BUTTON_1` or `BUTTON_LEFT` can be made into a Nim enum, but then the names (and usage of functions) will differ. On the other hand, having almost the same interface for all languages can make writing examples much easier, so I'm not sure.

We could look at it on a case by case basis. How would the usage change if those were enums? If it's just a matter of changing the capitalization and adding a bit of type safety that sounds ok, as long as the function signatures are more or less the same.

>     * I assume that wasm4 is a monorepo and so I'll have to commit all changes to the wrapper (although it's so simple that I think it won't have a lot of frequent changes) via the PRs? This is not an issue for me, so I'm just asking out of curiosity :)

Yup :smiley:

tjpalmer:
@Yardanico Are you still working this?

Also, gingerBill is working on this question of enum naming in Odin. I don't know the current state of the plans, though.

aduros:
@Yardanico Besides the snake tutorial is there any Nim documentation we're missing?

Yardanico:
@aduros No, apart from the snake tutorial all other examples have Nim code as well.

aduros:
Ok, let's consider that a separate issue and close this one :smiley:

PR #167, opened 2021-11-21
Initial Nim support

Yardanico:
This is the PR adding initial Nim support for wasm-4.

The first 4 points from https://github.com/aduros/wasm4/issues/147 are already done, most of the documentation except the Snake tutorial is also done.

I had to add two new template variables for `png2src` because in Nim the first element in an array has to be `uint8` for the whole array to also be `uint8`.

EDIT: The question about NimMain was resolved.


site/docs/guides/saving-data.md
```diff
@@ -37,6 +37,11 @@ var gameData int32 = 1337
 w4.DiskW(unsafe.Pointer(&gameData), unsafe.Sizeof(gameData))
 ```
 
+```nim
+var gameData = 1337'i32
+discard diskw(addr gameData, uint32 sizeof(gameData))
```
Yardanico, line 42: This is needed because Nim doesn't "throw away" results of functions if you don't use them - instead, you have to explicitly say `discard` to signify that you don't need the result.

aduros: Makes sense, thanks for the context!

Is the `uint32` part of `uint32 sizeof(gameData)` a cast? Would it make sense to change the type of that param in `diskw/diskr` to make the cast unnecessary?

aduros (approved):
I haven't set up Nim to play with this yet, but it looks excellent and very complete :smiley:  Thanks!

aduros:
> Nim allows global code in modules that will be executed on program start. Would it make sense to add [...] To the default Nim template so it works as expected by default?

I think that's probably the right call. If power users want to save on bytes they can remove it, but most language features should probably work out of the box.

site/docs/guides/saving-data.md
Yardanico: @aduros it's a type conversion which is safe (Nim does also have casts which look different). I can change the type and it'll still work because uint32 and Nim's `int` will both be 32-bit on wasm32, but on any other platform it'll be potentially incorrect

Yardanico:
@aduros there's still one Nim feature that doesn't correctly work - exceptions. Currently, Nim exits in some cases when it catches an exception, and `quit` requires a WASI function that WASM-4 doesn't have. I'll look into ways to solve it, but the current version sadly has that limitation.

Yardanico:
I'm resolving the merge conflicts now

site/docs/guides/saving-data.md
Yardanico: So what would be the way to go here?

aduros: Whatever you decide sounds good to me. I think we can always change it later if we really wanted to without breaking existing Nim projects.

aduros:
Really great work, thanks a lot! :tada: :tada: 

I'll set up Nim this weekend and take it for a spin, I'll let you know if anything comes up, but so far it looks really solid!

PR #171, opened 2021-11-23
Added Nyan Cat demo

jakeledoux:
Complete with sequenced music and bitmap animations!


jakeledoux:
Hmm... doesn't seem to play properly. I'll rescind this for now.

aduros:
It looks like a small typo: the cart `site/static/carts/nyancat` needs the `.wasm` filename suffix.

jakeledoux:
Whoops, don't know how I missed that.

aduros:
Very nice :cat: This is the best music in any cart so far! How did you implement it?

jakeledoux:
Thanks! It's nothing particularly inventive, I just stored the notes as integer offsets from C4 in different arrays for the two sections and calculated the frequencies at runtime. The drums are simply triggered by a modulo of the current time. [Here's the source.](https://github.com/jakeledoux/nyan/blob/main/src/lib.rs)

I've got a couple ideas of how to improve on this:

- You could fit more notes into each byte by choosing a key and only indexing within the scale
- Timed rests could save bytes (could also change the default step to 1/8th notes and denote smaller divisions with bit flags)
- Playback is currently tied to framerate which can get pretty slow, it would be good to find a way to decouple that

PR #174, opened 2021-11-26
Upgrade Minesweeper cartridge

claudiomattera:
WASM-4 function `seed()` was replaced with the count of frames from startup to first user input.

Cartridge was optimized for size (119 KiB to 20 KiB):

* all debug information are stripped;
* all unused functions, such as panic handlers and fmt module, are removed;
* custom font contains only used glyphs.


aduros:
Nice, thanks!

PR #175, opened 2021-11-28
feat: add wasm4-devtools

FaberVitale:
### Description

This PR adds a toggleable widget that displays helpful infos regarding the runtime state.
It is controlled via keyboard button (F8).

### Todo

- [x] show FPS
- [x]  display mouse position
- [x]  show active mouse buttons 
- [x]  display palette colors 
- [x] display pressed  gamepad buttons
- [x] display system flags
- [x] improve UI
- [x] add button/queryParam  to enable the widget on mobile devices
- ~~[ ] tweak update interval of `wasm4-devtools` widget~~
- [x] switch to a tab based layout
- [x] add keybindings panel
- [x] display all gamepads in a separate tab
- [x] minimize devtools
- [x] draw wasm memory
- [x] add `hex editor` view
- [x] figure out if the devtools should be rendered in `wasm4.org` and where the devtools should be rendered.
- [x] store pressed mouse  and gamepad buttons  (`storedGamepadByte & gamepadByteAfterUpdate`) to accurately display if a button  has been pressed between update intervals of the devtools.
- [x] Docs: Add Readme.md & typedocs
- [x] Docs: document theming customization via css variables.
- [x] add draw colors
- [x] add move `left| move right` buttons to change devtools position from right to left and viceversa
- [x] add `memoryMap` table

### Pics

Screenshot 2021-11-28 at 12 01 53

devtools-gamepad
Screenshot 2021-12-05 at 22 55 45


aduros:
Nice work!

Does this need to be in a separate module? It seems like things could be made simpler if it was in the web runtime module.

FaberVitale:
> Nice work!
> 
> Does this need to be in a separate module? It seems like things could be made simpler if it was in the web runtime module.

Unfortunately I do not agree with with this suggestion for the following reasons.

### 1. It not just a single devtools widget

I'm not just building a devtools widget but a collection of [custom elements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements) that can be reused outside the web runtime.
E.g. `<wasm4-gamepad>` is a custom element that shows a gamepage and it could be reused inside the `site/` .
Furthermore that package is already set up to be published to `npm` and it could easily distributed to be consumed by others.

### 2. Web runtime

Unfortunately web runtime is written in vanilla JS making the development full of hurdles and pitfalls caused by lack of strong typings and limited refactoring capabilities.
Futhermore using typescript enables meaningful intellisense in a modern IDE (VScode, IntelliJ).

I have no desire to develop this kind of project in vanilla JS.

### 3. The setup It is actually rather simple

When you run `npm install` inside `runtimes/web`  you first install then build ` devtools/web-devtools`.
I've set it up this way to simplify as much as possible the developer experience.

To be honest I think it is time to set up a proper js monorepo using either [pnpm](https://pnpm.io/) or [yarn 3](https://yarnpkg.com/getting-started).
In a proper monorepo the package manager symlinks the dependencies automatically during the install cmd:
the setup I've written would not be necessary.

### 4. Developing the components is rather easy

`web-devtools` uses [Vite](https://vitejs.dev/) to streamline and simplify the development.
Look at `vite.config.ts` it is really tiny compare to `runtimes/web/webpack.config.js`.

FaberVitale:
@aduros  should we render the devtools inside wasm4.org?

Currently the devtools are trapped inside the iframe and it doesn't look good.

Screenshot 2021-12-06 at 20 12 17

I think that there 2 possible choices:

1. disable devtools if we're inside an iframe.
2. mount `<wasm4-devtools></wasm4-devtools>` inside the parent document.

What do you think?

aduros:
For optimization can we include the devtools only in local development? So `w4 run/watch` would have the `F8` keybind for devtools, but not the `w4 bundle --html` output or games embedded on wasm4.org.

I think we would need to add another build output to the web runtime's webpack config called `wasm4-release.js` or similar?

FaberVitale:
> For optimization can we include the devtools only in local development? So `w4 run/watch` would have the `F8` keybind for devtools, but not the `w4 bundle --html` output or games embedded on wasm4.org.
> 
> I think we would need to add another build output to the web runtime's webpack config called `wasm4-release.js` or similar?

We can just use `webpack.DefinePlugin` to define a compile-time flag.
If done properly it would strip out devtools from the bundle if it is false.

FaberVitale:
> For optimization can we include the devtools only in local development? So `w4 run/watch` would have the `F8` keybind for devtools, but not the `w4 bundle --html` output or games embedded on wasm4.org.
> 
> I think we would need to add another build output to the web runtime's webpack config called `wasm4-release.js` or similar?

```
asset wasm4.js 79.9 KiB [emitted] [minimized] (name: main) (id hint: vendors) 1 related asset
asset wasm4.css 1.98 KiB [compared for emit] [minimized] (name: main) (id hint: vendors)
Entrypoint main 81.9 KiB = wasm4.css 1.98 KiB wasm4.js 79.9 KiB
orphan modules 758 KiB [orphan] 673 modules
runtime modules 670 bytes 3 modules
cacheable modules 219 KiB (javascript) 2.51 KiB (css/mini-extract)
  ./src/index.js + 8 modules 132 KiB [built] [code generated]
  ../../devtools/web/dist/web-devtools.es.js + 32 modules 86.9 KiB [built] [code generated]
  css ./node_modules/css-loader/dist/cjs.js!./src/styles.css 2.51 KiB [built] [code generated]
webpack 5.41.0 compiled successfully in 2732 ms

asset wasm4-release.js 23.4 KiB [emitted] [minimized] (name: main)
asset wasm4.css 1.98 KiB [compared for emit] [minimized] (name: main)
Entrypoint main 25.3 KiB = wasm4.css 1.98 KiB wasm4-release.js 23.4 KiB
orphan modules 116 KiB [orphan] 9 modules
cacheable modules 132 KiB (javascript) 2.51 KiB (css/mini-extract)
  ./src/index.js + 8 modules 132 KiB [built] [code generated]
  css ./node_modules/css-loader/dist/cjs.js!./src/styles.css 2.51 KiB [built] [code generated]
  ```

Done

aduros:
It's happening! :tada:

PR #177, opened 2021-11-29
Changed png2src template to mustache

christopher-kleine:
For more complex templates it makes sense to use a more feature-rich
template engine.
Since mustache was already a dependency, I converted
all templates to mustache.

I've also added the ability to output directly to a file instead to
Stdout.


aduros:
Nice!

Can you please also update this doc page with the new template syntax? https://wasm4.org/docs/guides/sprites#custom-template

PR #178, opened 2021-11-30
Added -q/--no-qr option to watch and run

christopher-kleine:
This option allows terminals that aren't good with QR codes to prevent the output of the QR code. It also provides a ENV var (W4_NO_QR) to set this on system/user level.


christopher-kleine:
I had some trouble with VSCode and Docker. Sometimes they would simple "hang". But only if the QR code was displayed.
Also: I tried it on a Windows maschine before I switch full-time to linux. And the PowerShell hat some problems as well. Even though I don't know if this still holds true.

DenialAdams:
To me "-q" means "--quiet", so if this ends up landing as is my suggestion would be not to include a short form. --no-qr isn't too long anyway

christopher-kleine:
I'm fine either way. I'll change it, since I agree with you here. At least about the `-q for --quiet` part.

aduros:
Can you please go back to showing the qr code by default, and only have a `--no-qr` option? (no `-q` short alias)

christopher-kleine:
Sure. Can do.

aduros:
Thanks! I'm going to make a few small tweaks.

PR #179, opened 2021-11-30
Providing reference to CLI

christopher-kleine:
A new page in the references to provide an overview of all possible commands of the `w4` CLI.

Each command is divided into those sections:

- Summary
- Usage
- Examples
- Table of options
- A more in-depth description of the command

The last point also includes setting up how to set default values (if possible).
I've also pointed out if the command has certain requirements (like png2src)


aduros:
Thanks :smile:

PR #183, opened 2021-12-06
[zig]: make string pointers null terminated

Apppppppple:
Zig has type-system support for null terminated pointers, but it isn't specified in the api right now.


aduros:
Good to know, thanks!

PR #184, opened 2021-12-06
fix(runtimes/web): runtime not paused when tab is in background

### LINKED ISSUE #172: Web Runtime doesn't pause when inactive
When playing on the website, once the game was started, it won't stop even if the canvas loses focus. This is not the case when switching to another tab however.

Steps to reproduce:

1. Open a cart like https://wasm4.org/play/nyancat
2. Click on the Play-Button
3. Switch to another tab
4. Return to the former tab
5. Click somewhere else on the site

Expected result:
- At step 3 the music should stop.
- At step 4 the music should continue.
- At step 5 the music should stop again.

Actual result:
- At step 3 the music stops. (expected)
- At step 4 the music continues. (expected)
- At step 5 the music continues. (not expected)

Side note: Sometimes the sound might actually play a note even when on a different tan.

https://user-images.githubusercontent.com/3718227/142984013-3aab8803-2df0-4d75-990d-1dcfa4460a12.mp4

FaberVitale:
Closes #172

Links:

- visibilitychange event: https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -71,9 +71,11 @@ async function loadCartWasm () {
 
         window.onblur = function () {
             document.body.classList.remove("focus");
+            runtime.updateIdleState();
         }
         window.onfocus = function () {
             document.body.classList.add("focus");
+            runtime.updateIdleState();
         }
     }
     document.body.classList.add("focus");
@@ -428,9 +430,6 @@ async function loadCartWasm () {
     window.addEventListener("pointerdown", onPointerEvent);
     window.addEventListener("pointermove", onPointerEvent);
     window.addEventListener("pointerup", onPointerEvent);
-
-    // @see https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event
-    window.addEventListener('visibilitychange', runtime.updatePausedFlag)
 
     const gamepadOverlay = document.getElementById("gamepad");
 
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -429,6 +429,9 @@ async function loadCartWasm () {
     window.addEventListener("pointermove", onPointerEvent);
     window.addEventListener("pointerup", onPointerEvent);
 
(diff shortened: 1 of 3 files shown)


aduros:
Hmm, this doesn't seem to resolve the issue for me. Here's how I'm testing:

1. Go to Nyan Cat and click play
2. Click somewhere else on the page to unfocus the game iframe and show the pause overlay
3. The music continues to play which indicates updates are still running. The music and updates should be paused when the pause overlay is shown

FaberVitale:
> Hmm, this doesn't seem to resolve the issue for me. Here's how I'm testing:
> 
>     1. Go to Nyan Cat and click play
> 
>     2. Click somewhere else on the page to unfocus the game iframe and show the pause overlay
> 
>     3. The music continues to play which indicates updates are still running. The music and updates should be paused when the pause overlay is shown



> Hmm, this doesn't seem to resolve the issue for me. Here's how I'm testing:
> 
>     1. Go to Nyan Cat and click play
> 
>     2. Click somewhere else on the page to unfocus the game iframe and show the pause overlay
> 
>     3. The music continues to play which indicates updates are still running. The music and updates should be paused when the pause overlay is shown

You're right.

Now on every window.blur/window.focus it runs after the 'focus' class is updated.

```js
class Runtime {
   [...]
    updateIdleState = () => {
        this.paused = !document.body.classList.contains('focus');

        if(this.paused) {
           this.pauseAudio();
        } else {
            this.unlockAudio();
        }
    }
}
````

I've tested a bit and I can't reproduce the bug.

aduros:
Thanks, that mutes the audio, but I don't think it addresses the root cause, which is that the cart is still receiving `update`s even when unfocused.

I can maybe take a stab at it this weekend. My pause handler code was kind of hacky and probably not easy to understand, but it could be simplified.

aduros:
Sorry I missed the part where you're actually toggling the pause. Looks good to me, thanks!

PR #190, opened 2021-12-09
Remove -nq from the watch command

Apppppppple:
This brings it in line with the run command


aduros:
Ah, I missed that one :smile: Thanks!

PR #192, opened 2021-12-11
Add zig version of snake tutorial

desttinghim:
What it says on the tin! I've mostly copy/pasted and changed what was different. I figured it was better to keep things similar than try to change things too much.

### DIFF
--- a/site/docs/tutorials/snake/collision-detection.md
+++ b/site/docs/tutorials/snake/collision-detection.md
@@ -17,6 +17,7 @@ Collision detection can be one of the harder to understand concepts of game deve
         {label: 'C / C++', value: 'language-cpp'},
         {label: 'Rust', value: 'language-rust'},
         {label: 'Go', value: 'language-go'},
+        {label: 'Zig', value: 'language-zig'},
     ]}>
 
 <TabItem value="language-typescript">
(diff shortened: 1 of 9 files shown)


site/docs/tutorials/snake/collision-detection.md
```diff
@@ -115,6 +116,49 @@ Now you're almost done. Only "Game Over" is left to finish this game.
 
 </TabItem>
 
+<TabItem value="language-typescript">
```
aduros, line 119:
```suggestion
<TabItem value="language-zig">
```

aduros (commented):
Thanks! This is looking great.

@christopher-kleine can you please take a look?

christopher-kleine:
I'm pretty clueless about Zig. So I can't help much here. But I noticed something else. Not related to Zig though.

site/docs/tutorials/snake/collision-detection.md
desttinghim: Oops, missed one. Thank you!

desttinghim:
@christopher-kleine What was it you noticed?

christopher-kleine:
Simple: In `Setup Your Project` there is a note titled `Prevent Browser from opening`.
The option `--no-browser` is not correct. It should be `--no-open`. Same with the env var `W4_NO_BROWSER`. It's `W4_NO_OPEN` instead.

But that should be my mistake. But I wouldn't mind if you were to fix this. More commits for you. 😆

desttinghim:
Thanks, I got that change committed as well. Might as well get it fixed.

aduros:
Thanks again for this!

PR #197, opened 2021-12-15
Stop when on trap instructions

### LINKED ISSUE #196: Handle traps in runtime
When a trap instruction happens it indicates that the wasm module has encountered an unrecoverable error. Currently the runtime will keep calling update in a loop when it encounters one, ignoring the error. If there is something broken, this generates a lot of error output. I think we should stop the runtime and possibly display an error to the user (though hopefully the errors will be gone at the point where you are sending it to users).

I think there's a way to get a stack trace too, but I haven't looked into this.

desttinghim:
I simply moved requestAnimationFrame to the end of the function. This
means the error will be encountered before the next frame is called.

Closes #196

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -442,7 +442,6 @@ async function loadCartWasm () {
 
     function loop () {
         processGamepad();
-        requestAnimationFrame(loop);
 
         const now = performance.now();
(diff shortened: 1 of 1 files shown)


aduros:
Nice, thanks!

PR #199, opened 2021-12-16
Adjust Nim template build code to show the issue more clearly when $WASI_SDK_ROOT is missing

XeroOl:
The default code to build nim has a more prominent error message when
attempting to build without WASI_SDK_ROOT defined.


aduros:
Nice, thanks!

PR #200, opened 2021-12-16
Replaced Tabs with DropDown Menu

christopher-kleine:
- Replaced all Tabs with DropDown Menus in Tutorial
- Added TODO to all missing languages
- Fixed Ordering of Rust/Go to Go/Nim/Odin/Rust
- Fixed a small mistake in Zig (Moving the Snake)
- Added Publish infos for Zig

### DIFF
--- a/site/docs/tutorials/snake/collision-detection.md
+++ b/site/docs/tutorials/snake/collision-detection.md
@@ -1,5 +1,4 @@
-import Tabs from '@theme/Tabs';
-import TabItem from '@theme/TabItem';
+import MultiLanguage, {Page} from '@site/src/components/MultiLanguage';
 
 # Collision Detection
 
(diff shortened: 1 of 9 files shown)


aduros:
It's looking great! :smile: 

Is the `// TODO` text supposed to be displayed on the page? Right now if you select Odin the content is empty, which seems like something is broken.

christopher-kleine:
Yes, this is supposed to show up. If I don't use a placeholder, it's simply empty. This at least shows that the content will be added in the future instead of a blank slate.

aduros:
Yeah, but it looks like the TODO is not actually displayed on the website. Maybe it's being interpreted as a JS comment by docusaurus?

christopher-kleine:
No, that was my mistake. I simply forgot it the "simple" cases. I've updated this and also fixed some other occurrences.

aduros:
Ok,  thanks!

Issue #198, opened 2021-12-16
Dynamic Port to allow multiple instances running simultaneously


XeroOl:
Currently, if you try to use `w4 run` in multiple terminals at the same time, only one will work, and the rest will crash because the port is already in use. It might be better to choose a port number dynamically, to allow multiple instances to run concurrently.

PR #202, opened 2021-12-16
Adding custom ports to run and watch

christopher-kleine:
The new `port` option allows to specify one or more ranges for the port.

The default is a range between 4000 and 6000

A range can be specified like `4000-6000`.

A fixed port can also be specified: --port 4444

It's also possible to provide multiple ranges:
`--port 4000-6000,8000-7000,8080`

If a port is taken, it attempts to do this again.
After 10 attempts, an exception is thrown.

This Solves #198

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -45,6 +45,11 @@ program.command("watch")
     .description("Rebuild and refresh when source code changes")
     .option("--open", "Open the browser", true)
     .addOption(
+        new Option("--port <port-range>", "Binds the runtime to a specific range of ports")
+            .env("W4_PORT")
+            .default("4000-6000")
+    )
+    .addOption(
         new Option("-n, --no-open", "Don't open the browser")
             .env("W4_NO_OPEN")
             .default(false)
@@ -63,6 +68,11 @@ program.command("watch")
 program.command("run <cart>")
     .description("Open a cartridge in the web runtime")
     .option("--open", "Open the browser", true)
+    .addOption(
+        new Option("--port <port-range>", "Binds the runtime to a specific range of ports")
+            .env("W4_PORT")
+            .default("4000-6000")
+    )
     .addOption(
         new Option("-n, --no-open", "Don't open the browser")
             .env("W4_NO_OPEN")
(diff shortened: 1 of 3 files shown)


aduros:
Thanks for this! I have some suggestions:

- Instead of printing every failed port attempt, it should only print once if an alternate port was successfully used. eg: `Using port 4450, since port 4444 is already in use`.
- Let's simplify the `--port` option to just take a single port number... I can't think of a situation where users would want a custom range, or a list of ranges.
- Port attempts should be sequential instead of random. It should first try 4444, then 4445, etc. We can abort after 100 attempts instead of 10 (with a friendly error message asking to supply a different --port).
- The default port should remain 4444.

christopher-kleine:
Let me summarize this:

```bash
w4 run                  # Using Port 4444
# Output: Listening on Port 4444

w4 run --port 4445      # Using Port 4445
# Output: Listening on Port 4445

w4 run --port 4444-4500 # Invalid syntax
# Output: Parse error on property "port"

w4 run --port 4444,4445 # Invalid syntax
# Output: Parse error on property "port"

w4 run --port 80        # Using privileged Port 80; Port 1025 is the first without privileges
# Output: Unable to bind on port 80.
# Output: Using 1025 instead
# Output: Listening on Port 1025

w4 run --port 1        # Using privileged Port 1; Port 1025 would the first without privileges
# Output: Unable to bind on port 1.
# Output: Error: Unable to find available port. Please specify a different port.
```

Is this correct?

joshgoebel:
Looks good to me, though I wonder if we should avoid trying to bind on system ports to begin with and just skip them?

christopher-kleine:
If you run this as root or in a docker environment, system ports are perfectly valid.
Also: I only used them in this example because I kinda *know* they would fail if you're running it as an unprivileged user. At least on Linux.

joshgoebel:
> If you run this as root 

Probably a bad idea, and what I was thinking of.

> or in a docker environment, system ports are perfectly valid.

That's a fair point. :)

aduros:
> Is this correct?

Yes, that looks perfect!

christopher-kleine:
I almost forgot: I've added a more "speaking" explanation if port binding fails. If this is too much (or not enough), let me know.

aduros:
Wonderful, thanks a lot!

PR #208, opened 2021-12-18
Add maze game

binji:


aduros:
Amazing fog effect with the dithering! :exploding_head: Very awesome effect.

Issue #204, opened 2021-12-16
Nelua support


christopher-kleine:
[Nelua](https://nelua.io/) can compile to WebAssembly using [Emscripten](https://emscripten.org/). An example can be found here: https://github.com/edubart/nelua-game2048

- [x] `w4 new`: Add new project template
- [x] `w4 watch`: Add new filename suffix + build/run support
- [x] `w4 png2src`: Add new language parameter
- [x] Website: Setup documentation
- [x] Website: Translated code snippets for all guide pages
- [x] Website: Add code to Snake tutorial

Andre-LA:
I'm investigating this at my free time, I'll give an update (maybe on discord instead) once I got some progress.

aduros:
I made some progress with this and have a minimal cart compiling and running:

`build.sh`:
```shell
CFLAGS="-mexec-model=reactor -Wl,--allow-undefined,-zstack-size=1024,--no-entry,--import-memory,--initial-memory=65536,--max-memory=65536,--global-base=6560,--export-dynamic"

nelua \
    --cc "$WASI_SDK_PATH/bin/clang --sysroot=$WASI_SDK_PATH/share/wasi-sysroot" \
    --cflags="$CFLAGS" --verbose \
    ./wasm4-test.nelua -H -o out.wasm --release
```

`wasm4-test.nelua`:
```lua
-- Pointer to WASM-4 memory register
local DRAW_COLORS = (@*uint16)(0x14);

-- Import some WASM-4 functions
local function text (str: cstring <const>, x: int32, y: int32): void <cimport> end
local function trace (str: cstring <const>): void <cimport> end

local x = 69

local function update (): void <cexport, codename "update">
    -- $ is used to deref a pointer
    $DRAW_COLORS = 0x23

    -- Call an imported function
    text("Hello from Nelua!", x, 20)
    x = (x + 1) % 160

    trace("Trace works")
end

-- BUG: This doesn't get printed or even included in the .wasm. Need to look more into it, make sure
-- main() is being exported?
trace("Hello from initializer")
```

Some possible next steps:

- Create a wasm4.nelua with all the bindings
- Nelua lets you add compiler configuration in source files like this: https://github.com/edubart/nelua-game2048/blob/master/config.nelua. I think we should be able to move all the clang flags into the top of wasm4.nelua so we don't need a build.sh?
- Figure out how to make root level code blocks work (the trace bug above)
- I haven't tested GC or memory allocation at all yet

[...]

PR #210, opened 2021-12-18
add function bindings to nelua, move cflags from Makefile to wasm4.nelua (#204)

AKDev21:
additionally,

- add .gitattributes for using lua as syntax highlighting on github for .nelua files
- add ``--verbose`` flag in debug builds


aduros:
Thanks! :moon:

PR #218, opened 2021-12-21
(fix) HTML runtime template

joshgoebel:
Fix for to bring into alignment with d032fd69dafa37944c86d9f001fcf731f4f320d4


aduros:
Nice catch, thanks!

PR #224, opened 2021-12-22
add incomplete nelua docs

AKDev21:


AKDev21:
looks like something is missing for "changing the language on website to nelua" to work.

i would love if you anyone can fix that because i don't have experience with working with node.js, javscript and Docusaurus.

Thanks :smiley_cat:

aduros:
Thanks a lot for tackling this! :smile: 

I just added the Nelua option to the site, if you want to update this branch with the latest it should show up. Please also replace `nelua` with `lua` as the markdown code language so that we get syntax highlighting. We can add true Nelua syntax highlighting later but it's probably a larger task.

aduros:
Also please add the Nelua sections in alphabetical order, so between `go` and `nim' :nerd_face:

AKDev21:
Done :smile_cat:

PR #234, opened 2021-12-23
Nelua improvements

Andre-LA:
This PR improves the nelua bindings and add the remaining documentation for the website, it's related with #204 

However, I also have a question, the bindings uses [`<comptime>` annotations](https://nelua.io/overview/#compile-time-variables), like this:
```lua
global MOUSE_LEFT <comptime> = 1
global MOUSE_RIGHT <comptime> = 2
global MOUSE_MIDDLE <comptime> = 4
```
This way, the nelua compiler generates more efficient code.

My question is, should `png2src` also emit `<comptime>` constants instead of run-time ones? Like this:
```lua
local bunny_width <comptime> = 16
local bunny_height <comptime> = 16
local bunny_flags <comptime> = 1 -- BLIT_2BPP
local bunny: [64]uint8 <const> = { --[[...]] }
```

If so, I can commit and add this change in this PR if you wish.

### DIFF
--- a/cli/assets/templates/nelua/src/main.nelua
+++ b/cli/assets/templates/nelua/src/main.nelua
@@ -1,25 +1,27 @@
 require "wasm4"
 
 local smiley: []uint8 = {
-    0b11000011,
-    0b10000001,
-    0b00100100,
-    0b00100100,
-    0b00000000,
-    0b00100100,
-    0b10011001,
-    0b11000011,
+  0b11000011,
+  0b10000001,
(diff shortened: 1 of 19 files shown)


AKDev21:
review: everything looks fine.

AKDev21 (approved):

aduros:
This looks great, thanks a lot!

> My question is, should png2src also emit <comptime> constants instead of run-time ones?

That probably makes sense, I'll update it.

Issue #228, opened 2021-12-22
`blit` function not working correctly with small sprites  


luehmann:
It seems `blit` doesn't work for sprites smaller than 8x8.

@aduros on Discord:
> I could be wrong but there may be a limitation where the width of a sprite needs to be on a byte boundary. So the lowest width you can have of a 2bpp sprite is 4 and the lowest for a 1bpp is 8.

luehmann:
fixed in #236

PR #236, opened 2021-12-23
fix blit for non-byte-aligned sprites

luehmann:
After investigating #228 I found that `x` was used to calculate the `shift`. I believe that using the `bitIndex` instead should result in the actual expected behavior. 

- [x] fix in web runtime
- [x] fix in native runtime

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -348,14 +348,15 @@ void w4_framebufferBlit (const uint8_t* sprite, int dstX, int dstY, int width, i
             // Sample the sprite to get a color index
             int colorIdx;
             int x = srcX + sx, y = srcY + sy;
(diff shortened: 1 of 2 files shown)


joshgoebel:
This fix looks like it would apply to any sprite that wasn't an even multiple of 8 wide, not just small sprites...

luehmann:
> This fix looks like it would apply to any sprite that wasn't an even multiple of 8 wide, not just small sprites...

Yeah well, it does apply to all spites, but I've created a [example card](https://github.com/luehmann/wasm4-test-cart) and everything seems to be fine.

joshgoebel:
Yeah, it looks ok at a glance I was just nothing that the issue description is incorrect.

joshgoebel (approved):

aduros:
Seems to work nicely, thanks a lot!

Issue #235, opened 2021-12-23
run-native: stack overflow ('_initialize')


sporksmith:
[cart-debug.wasm.gz](https://github.com/aduros/wasm4/files/7771670/cart-debug.wasm.gz) works as expected in the browser via `w4 run`, but fails using the native backend `w4 run-native`:

```
$ w4 --version
2.0.0
$ w4 run-native ./cart-debug.wasm 
WASM error: [trap] stack overflow ('_initialize')
```

I get the same error using the wasm4 native backend built at @ 85a40001ad0e4691b01715a8f76f9ee830af22db.

The problem goes away when I turn on any optimizations compiling my cartridge with zig. e.g. with `-Drelease-safe=true`: [cart-safe.wasm.gz](https://github.com/aduros/wasm4/files/7771669/cart-safe.wasm.gz)

One guess is that the code section of the debug build is simply too large, since wasm-objdump shows it taking ~100K.

```
$ wasm-objdump -h cart-debug.wasm

cart-debug.wasm:	file format wasm 0x1

Sections:

     Type start=0x0000000b end=0x00000193 (size=0x00000188) count: 62
   Import start=0x00000195 end=0x000001c7 (size=0x00000032) count: 4
 Function start=0x000001ca end=0x00000359 (size=0x0000018f) count: 397
    Table start=0x0000035b end=0x00000360 (size=0x00000005) count: 1
   Global start=0x00000362 end=0x0000038e (size=0x0000002c) count: 7
   Export start=0x00000391 end=0x00001170 (size=0x00000ddf) count: 244
     Elem start=0x00001172 end=0x0000117c (size=0x0000000a) count: 1
     Code start=0x00001180 end=0x00019ac0 (size=0x00018940) count: 397
     Data start=0x00019ac3 end=0x0001d74a (size=0x00003c87) count: 3
   Custom start=0x0001d74d end=0x0001f92f (size=0x000021e2) "name"
   Custom start=0x0001f931 end=0x0001f94b (size=0x0000001a) "producers"
```

If that's the issue, though, I'm not sure why it *does* work in the browser back-end, and it'd be nice if there were a clearer error message. **EDIT**: this isn't the issue. This cart does exceed the 64 KB cart size limit, but wasm doesn't load code into linear memory, and the cart size limit appears to currently be unenforced.

The cart itself is built from https://github.com/sporksmith/snake-wasm4

sporksmith:
[...]

PR #240, opened 2021-12-24
Increase wasm3 stack size from 4 KB to 64 KB.

sporksmith:
This is an arbitrary limit corresponding to the implementation details
of the wasm3 interpreter. It's unrelated to the resource constraints of
the wasm4 VM. Making this too small will ultimately result in `[trap]
stack overflow` at the command line.

Using 64 KB, since this is the default of wasm3 standalone binary on
desktop platforms (from wasm3/platforms/app/main.c).

Fixes https://github.com/aduros/wasm4/issues/235

### DIFF
--- a/runtimes/native/src/backend/wasm_wasm3.c
+++ b/runtimes/native/src/backend/wasm_wasm3.c
@@ -166,7 +166,17 @@ static void check (M3Result result) {
 
 uint8_t* w4_wasmInit () {
     env = m3_NewEnvironment();
-    runtime = m3_NewRuntime(env, 4096, NULL);
+
+    // This is an arbitrary limit corresponding to the implementation details
+    // of the wasm3 interpreter. It's unrelated to the resource constraints of
+    // the wasm4 VM. Making this too small will ultimately result in `[trap]
+    // stack overflow` at the command line.
+    //
+    // Using 64 KB, since this is the default of wasm3 standalone binary on
+    // desktop platforms (from wasm3/platforms/app/main.c).
+    uint32_t wasm3StackSize = 64 * 1024;
+
+    runtime = m3_NewRuntime(env, wasm3StackSize, NULL);
 
     runtime->memory.maxPages = 1;
     ResizeMemory(runtime, 1);


sporksmith:
Dug into this some more together with @binji on Discord.

This is the [WebAssembly native stack](https://webassembly.github.io/spec/core/exec/runtime.html#stack). Deciding the limit here is a bit tricky.

Theoretically the native stack could be abused by wasm4 cartridges for "free" memory, e.g. using recursion. Having some smallish limit here could mitigate that. OTOH there is no API to limit the native stack usage in the browser wasm engines.
Also, different backends could use different amounts of stack when executing the same wasm - this isn't specified by the wasm spec.

Some options:

* Bump up the limit to something relatively large as in the current version of this PR. That'd put the native backend on more even footing with the web backend. People might get clever and abuse the native stack as extra memory, but it's not straightforward to do.
* Keep a relatively small limit here, and make "X bytes of native stack on the w4 native backend" part of the w4 VM spec. This is a bit unsatisfying since the stack usage may also depend what platform the w4 native backend is compiled for, compilation options, etc. A cartridge that fits on the native backend on one platform might no longer fit on the next release of w4 or on a different platform.
* Expose the native stack size as a command-line option to `w4 run-native`. In that case we could potentially keep a relatively small default here, but at least folks would have an "escape valve"
* Make the "out of native stack" error message a bit friendlier and include some pointers to what it means and how it might be avoided. In particular it might be good to point out somewhere that zig debug builds use quite a lot of native stack.

sporksmith:
I'm not sure, but I *think* forbidding recursion might close the loophole of using the native stack for extra memory. It'd require some extra static analysis of the wasm to enforce, but I think it might not be too difficult. OTOH it'd feel like a pretty artificial restriction, and could get violated unintentionally, and by library code, etc.

joshgoebel:
Curious: Did you try anything between 4kb and 64kb to see where it starts working for your sample case?

sporksmith:
@joshgoebel 
> Curious: Did you try anything between 4kb and 64kb to see where it starts working for your sample case?

8 KB also worked. Given that this was mostly just the snake demo though, I suspect a 8 KB limit would also be hit fairly easily, at least in zig debug builds.

aduros:
Makes perfect sense, thanks for the PR!

We might need to bump this up further, or make it configurable somehow, but for now this looks good to me.

PR #243, opened 2021-12-26
[Docs][Wasm4] optionally set initial language of MultiLanguageCode from query params

FaberVitale:
This PR adds the option to set the initial language from query params, param `code-lang`, and fixes several react related issues in `MultiLanguage.js` & `MultiLanguageCode.js`

example: https://deploy-preview-243--wasm4.netlify.app/docs/guides/saving-data/?code-lang=go#reading-data-from-disk

### DIFF
--- a/site/src/components/MultiLanguage.js
+++ b/site/src/components/MultiLanguage.js
@@ -15,6 +15,20 @@ const names = {
     "zig": "Zig",
 };
 
+/**
+ * @type {(val: unknown) => val is keyof typeof names}
+ */
+function isValidLanguageValue(value) {
+    return typeof value === 'string' && Object.prototype.hasOwnProperty.call(names, value);
+}
+
+/**
+ * @type {(val: string | null | undefined) => string}
+ */
+function normalizeLanguageValue(val) {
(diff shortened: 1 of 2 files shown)


aduros:
Thank you for fixing those React warnings :smile: 

Currently on live if you change the language of one code box, it will also change the language of all other code boxes on the current page. That seems to not work anymore in this PR.

Also do you know how we would use this feature, or make it discoverable in some way?

FaberVitale:
> Also do you know how we would use this feature, or make it discoverable in some way?

### Create new pages

In terms of SEO it would better to have languages as route params:
e.g. `/docs/guides/saving-data/go`.

You also have to link them inside `/docs/guides/saving-data`.

Unfortunately Docusaurus is not as flexible as `next.js`:

1.  you need to write a custom plugin that handles these new routes
2.  you need to manage the client side routing via react-router.

I've found a useful answer on stack overflow: https://stackoverflow.com/a/63906923

---

If we want to keep on using the queryParams we could just update  `code-lang` query param on Multilanguage value change.
Note that even in this case it would make sense make these options available as links.

aduros:
Awesome, thanks!

PR #249, opened 2021-12-27
(enh) add blue screen for WASM Runtime errors

### LINKED ISSUE #245: Add blue screen of death for WASM crashes
When there is a known common WASM crash, we should indicate it visibly.  Memory out of bounds errors seem to be quite common in my development so far.  If there were other common ones we could handle those as well.  I'm happy to work on this if we could agree on wording and visual style.

Mockup

Screen Shot 2021-12-26 at 4 05 40 PM
:

joshgoebel:
Adds blue screen error handling for common WASM Runtime errors.

Resolves #245 

- [x] web runtime
- [x] <del>native runtime</del> not necessary per @aduros 

Screen Shot 2021-12-27 at 8 35 10 AM
Screen Shot 2021-12-27 at 9 07 01 AM

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -473,7 +473,9 @@ async function loadCartWasm () {
 
         if (deltaFrame >= INTERVAL) {
             lastFrame = now - (deltaFrame % INTERVAL);
-            runtime.update();
+            if (!runtime.crashed) {
+                runtime.update();
+            }
 
             gamepadOverlay.style.display = runtime.getSystemFlag(constants.SYSTEM_HIDE_GAMEPAD_OVERLAY)
                 ? "none" : "";
--- a/runtimes/web/src/runtime.js
+++ b/runtimes/web/src/runtime.js
@@ -29,6 +29,7 @@ export class Runtime {
         this.reset();
 
         this.paused = false;
+        this.crashed = false;
     }
 
     setMouse (x, y, buttons) {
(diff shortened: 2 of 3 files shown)


runtimes/web/src/index.js
```diff
@@ -473,7 +473,9 @@ async function loadCartWasm () {
 
         if (deltaFrame >= INTERVAL) {
             lastFrame = now - (deltaFrame % INTERVAL);
-            runtime.update();
+            if (!runtime.crashed) {
+                runtime.update();
+            }
```
joshgoebel, line 478: This doesn't actually STOP the requestAnimation loop, but I think that's ok since it essentially turns it into just a NOP.

joshgoebel:
Let me know how this looks and if it's good I can see about the native runtime.

runtimes/web/src/runtime.js
```diff
@@ -289,12 +291,71 @@ export class Runtime {
...
+            message = "This cartridge has\nattempted a memory\naccess that is\nout of bounds.";
+        }
+        message += "\n\n\n\n\nHit R to reboot.";
+        return message;
+    }
+
+    blueScreen(err) {
+        this.crashed = true;
+
+        const COLORS = [
```
joshgoebel, line 326: I could move this into `config` as CRASH_PALETTE also...

aduros:
Thanks, I'll take a look!

Let's implement this for the web runtime only. On native we can just crash and print something to stdout.

joshgoebel:
> Let's implement this for the web runtime only.

Sweet! No argument here!

aduros:
Thanks!

PR #256, opened 2021-12-28
Expose tracef in zig template

### LINKED ISSUE #244: (zig) Add `tracef` API call  (APIs should have feature parity)
The C API has `tracef`, so should Zig.

sporksmith:
Fixes #244

### DIFF
--- a/cli/assets/templates/zig/src/wasm4.zig
+++ b/cli/assets/templates/zig/src/wasm4.zig
@@ -121,3 +121,7 @@ pub fn trace(x: []const u8) void {
     traceUtf8(x.ptr, x.len);
 }
 extern fn traceUtf8(strPtr: [*]const u8, strLen: usize) void;
+/// Use with caution, as there's no compile-time type checking.
+/// %d and %f expects 32-bit integers and floats, respectively.
+/// See https://github.com/aduros/wasm4/issues/244.
+pub extern fn tracef(x: [*:0]const u8, ...) void;
--- a/cli/assets/templates/zig/src/wasm4.zig
+++ b/cli/assets/templates/zig/src/wasm4.zig
@@ -121,7 +121,13 @@ pub fn trace(x: []const u8) void {
     traceUtf8(x.ptr, x.len);
 }
 extern fn traceUtf8(strPtr: [*]const u8, strLen: usize) void;
+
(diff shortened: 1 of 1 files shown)


sporksmith:
This is the "light" solution. I'm going to send an alternate PR shortly with a safer and more idiomatic interface, but it requires a bit more code.

sporksmith:
> This is the "light" solution. I'm going to send an alternate PR shortly with a safer and more idiomatic interface, but it requires a bit more code.

After further consideration, I'm just proposing this PR for now. I think the safe wrapper around `tracef` is too complex to belong here. (One minor snag I ran into is it doesn't appear possible to both declare an `extern` `tracef` and provide a public function named `tracef` in the same file)

sporksmith:
As a point of comparison, this is pretty much how the zig standard library expose's libc's `printf`: https://github.com/ziglang/zig/blob/4f4f0bc6f03a84597247504a5ee15ed7f1e050d1/lib/std/c.zig#L105

joshgoebel:
>  it doesn't appear possible to both declare an extern tracef and provide a public function named tracef in the same file)

Yeah I was worried about that.

> As a point of comparison, this is pretty much how the zig standard library expose's libc's printf:

Then I'd say we're in good company. :)

cli/assets/templates/zig/src/wasm4.zig
```diff
@@ -121,3 +121,7 @@ pub fn trace(x: []const u8) void {
     traceUtf8(x.ptr, x.len);
 }
 extern fn traceUtf8(strPtr: [*]const u8, strLen: usize) void;
+/// Use with caution, as there's no compile-time type checking.
+/// %d and %f expects 32-bit integers and floats, respectively.
```
joshgoebel, line 125: I thought I commented, but now I don't see it.  Are you certain?  Doesn't it take `f64` not `f32`??? We move the pointer 8 bytes for floats in the runtime code.

sporksmith: Oops, you're right!

sporksmith: Fixed

joshgoebel (approved):
Much better!

aduros:
Thanks!

Issue #185, opened 2021-12-08
Add keyboard controls for Gamepad 2


aduros:
In order to enable multiplayer games, we should have a way to input for player 2.

Maybe copy Pico's keybinds? Keys ESDF for directionals, QW for action buttons.

Ideally this should work on European keyboard layouts (and maybe even Dvorak).

desttinghim:
I only know US keyboards, but what about WASD and RT? WASD is a pretty common setup for controlling movement, and RT are nearby. I have no idea if it would map well to other layouts. Would it be difficult to make the controls remappable?

aduros:
R is currently mapped to reboot the cart, although we can always change the existing mappings.

One benefit of ESDF is that it works with [French AZERTY keyboards](https://en.wikipedia.org/wiki/AZERTY). I was wrong about QW being used for the buttons in Pico, they actually use LShift or Tab for button 1, and A or Q for button 2.

For now I think we should borrow [Pico's mappings](https://pico-8.fandom.com/wiki/Btn) with ESDF + LShift/Tab + A/Q for Player 2. Also adding N + M keys for Player 1 so that all four hands can comfortably fit on the keyboard :smile:

Making the controls remappable in the future would be nice, but we'll need to design the UI for it.

PR #263, opened 2021-12-30
Add cartridge for Racer

claudiomattera:
Here it comes a full-fledged racing game :)

Hear the engine rumbling as you accelerate, swing left and right enjoying the distant view across three different tracks, and enter your old-fashioned three-letters signature if you set a high score!

The main core of the game was written following the video [Code-It-Yourself! Retro Arcade Racing Game] (great content, by the way).
I implemented it from scratch for WASM-4, and then I added a few more details.

[Code-It-Yourself! Retro Arcade Racing Game]: https://www.youtube.com/watch?v=KkMZI5Jbf18

It was mostly meant as a laboratory to experiment and to develop and improve some libraries I plan to use for other games, but it turned out to be quite playable, too.
There are countless aspects that I could improve (*especially* backgrounds and palettes), but I felt like I had to publish it now, otherwise I would never finish :)

Bonus content: the game features multiplayer races in split screen.
I tested it by adding random noise to the first gamepad so that each car would behave slightly different, and I look forward to see support for multiple gamepads.


aduros:
Absolutely incredible game!! Awesome graphics and the engine sound effects are :fire: 

Could you move the game to the top of the games list? In the future (post gamejam?) we might redesign the games page and allow grouping by author, but for now I want to make sure it gets some visibility as it's a great demo.

BTW, keyboard input for gamepad 2 was implemented in #185. It seems to be working well :grinning:

claudiomattera:
I moved the game at the top of the list.
I also took the chance to make a last-minute change in the palette in the second track, I hope it looks like a sunset now.

PR #266, opened 2021-12-31
zig template: export only the 'start' and 'update' functions

sporksmith:
Workaround for https://github.com/ziglang/zig/issues/2910.

See in particular:
https://github.com/ziglang/zig/issues/2910#issuecomment-1003455815
https://github.com/ziglang/zig/issues/2910#issuecomment-1003456922


aduros:
Thanks!

PR #269, opened 2022-01-03
Use packed structs rather than bit manipulations in zig

pfgithub:
This

- makes it nicer to write code for wasm4 in zig

but also

- makes the language template more complicated (122 lines → 202 lines, now defines some structs and is no longer just a bunch of extern fns)
- makes zig examples different from examples in other languages, making it harder to update documentation.

While I'm personally using this in my project, I'm not sure if this would be good as the default template or not.

This lets you do:

```zig
    if(w4.MOUSE.buttons.left) {
```

instead of

```zig
    if(w4.MOUSE & w4.MOUSE_LEFT != 0) {
```

and it lets you write

```zig
w4.tone(
    .{ .start = 262, .end = 523 },
    .{ .sustain = 60, .release = 30 },
    100,
    .{ .channel = .pulse1, .mode = .p50 },
);
```

instead of

```zig
w4.tone(
    262 | (523 << 16),
    60 | (30 << 8),
    100,
    w4.TONE_PULSE1 | w4.TONE_MODE3,
);
```

### DIFF
--- a/cli/assets/templates/zig/src/wasm4.zig
+++ b/cli/assets/templates/zig/src/wasm4.zig
@@ -86,17 +86,56 @@ pub extern fn hline(x: i32, y: i32, len: u32) void;
 // │                                                                           │
 // └───────────────────────────────────────────────────────────────────────────┘
 
-/// Plays a sound tone.
-pub extern fn tone(frequency: u32, duration: u32, volume: u32, flags: u32) void;
+const externs = struct {
+    extern fn tone(frequency: u32, duration: u32, volume: u32, flags: u32) void;
+};
 
-pub const TONE_PULSE1: u32 = 0;
-pub const TONE_PULSE2: u32 = 1;
-pub const TONE_TRIANGLE: u32 = 2;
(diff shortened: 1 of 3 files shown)


pfgithub:
Nevermind, this would have to mess with the reference page because it makes things too different between languages

aduros:
Yeah, the intent with the base API is to be as simple as possible. I expect and encourage that users create more fluent libraries on top of the base system though :smiley: 

For example, this [Rust library](https://crates.io/crates/wasm4) adds more Rust-friendly interfaces. Something similar for Zig could definitely be useful.

joshgoebel:
> I expect and encourage that users create more fluent libraries

@pfgithub Please create a new repository for this and I'd love to perhaps contribute a little over time.  I'm going to borrow a lot of these ideas for the TIC-80 Zig library I'm working on also.

pfgithub:
@joshgoebel Here: https://github.com/pfgithub/wasm4-zig

I also have some other changes in my own project like using `std.meta.Vector(2, i32)` for positions although I'm not sure if that's a good way to use zig.

Issue #270, opened 2022-01-04
Example pixel functions fail to subtract 1 from draw color


sporksmith:
The `DRAW_COLOR`s are numbered 1-4, but are represented in the `FRAMEBUFFER` with 2 bits as 0-3. The example pixel functions fail to take this into account. e.g. from https://wasm4.org/docs/guides/basic-drawing/?code-lang=assemblyscript#direct-framebuffer-access :

```
// Use the first DRAW_COLOR as the pixel color
const color = u8(load<u16>(w4.DRAW_COLORS) & 0b11);
// Write to the framebuffer
store<u8>(w4.FRAMEBUFFER + idx, (color << shift) | (load<u8>(w4.FRAMEBUFFER + idx) & ~mask));
```

The documentation is also a bit light on this in general, though this behavior is implied by 
 e.g. "[Each draw color can be a value between 1 and 4 representing a palette color, or 0 to signify transparency.](https://wasm4.org/docs/reference/memory#draw_colors)".

Test program that demonstrates the mapping (in zig):

```
export fn update() void {
    w4.tracef("before: %x", @as(u32, w4.FRAMEBUFFER[0]));

    var color: u8 = 0;
    while (color <= 4) : (color += 1) {
        w4.DRAW_COLORS.* = color;
        w4.rect(0, 0, 1, 1);
        w4.tracef("after rect with DRAW_COLORS=%d, FRAMEBUFFER[0]=%d", @as(u32, w4.DRAW_COLORS.*), @as(u32, w4.FRAMEBUFFER[0]));
    }
}
```

Produces output:

```
before: 0
after rect with DRAW_COLORS=0, FRAMEBUFFER[0]=0
after rect with DRAW_COLORS=1, FRAMEBUFFER[0]=0
after rect with DRAW_COLORS=2, FRAMEBUFFER[0]=1
after rect with DRAW_COLORS=3, FRAMEBUFFER[0]=2
after rect with DRAW_COLORS=4, FRAMEBUFFER[0]=3
```

joshgoebel:
This is correct.  The frame buffer (per pixels) is 2-bit it can only hold 0 1 2 3.

The 1-4 only exists at the API level.  I assume it's to make it "Easier" for some people, but it is confusing.

sporksmith:
Which `this` are you referring to? I'm not suggesting a bug in `w4` itself, just in the example `pixel` functions on the website.

joshgoebel:
[...]

PR #271, opened 2022-01-04
Account for palette colors 1-4 being represented as 0-3 in frame buffer

sporksmith:
Fixes https://github.com/aduros/wasm4/issues/270

### DIFF
--- a/site/docs/guides/basic-drawing.md
+++ b/site/docs/guides/basic-drawing.md
@@ -101,7 +101,7 @@ The first color in the palette register is used as the screen background color.
 
 ## The `DRAW_COLORS` Register
 
-All drawing functions are affected by the `DRAW_COLORS` memory register. `DRAW_COLORS` is a 16 bit value that can store up to 4 colors.
+All drawing functions are affected by the `DRAW_COLORS` memory register. `DRAW_COLORS` is a 16 bit value that can store up to 4 colors, using 4 bits each.
 
 For example, `rect()` uses the first draw color for the fill color, and the
 second draw color as the outline color. To draw a light-green (palette color 2)
@@ -235,7 +235,9 @@ For info on other shape drawing functions like `line()` and `oval()`, see the [F
 ## Direct Framebuffer Access
 
 The `FRAMEBUFFER` memory region contains the framebuffer, with each byte containing 4 pixels (2 bits
-per pixel). For example, to clear the entire screen to palette color 3:
+per pixel). In the framebuffer, the palette colors 1-4 are represented numerically as 0-3.
+
+For example, to clear the entire screen to palette color 4, we write 3 into each position:
 
 <MultiLanguageCode>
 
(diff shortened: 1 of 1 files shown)


site/docs/guides/basic-drawing.md
```diff
@@ -484,11 +522,12 @@ fn pixel(x: i32, y: i32) {
...
 
   ;; Use the first DRAW_COLOR as the pixel color
-  ;; color = *DRAW_COLORS & 0b11;
+  ;; color = *DRAW_COLORS & 0b1111;
   (local.set $color
     (i32.and
       (i32.load16_u (global.get $DRAW_COLORS))
-      (i32.const 3)))
+      (i32.const 15)))
+  ;; FIXME: return if $color is zero, then subtract 1.
```
binji, line 530:
```suggestion
  ;; return if $color is zero, then subtract 1.
  (br_if (i32.eqz (local.get $color)))
  (local.set $color (i32.sub (local.get $color) (i32.const 1)))
```

binji: This is the most natural translation of the other code, though you could also subtract 1 ahead of time, and exit the function if the color is < 0

sporksmith:
Actually these should probably all also mask again after subtracting. Otherwise a 0xf in the palette color could result in a 4-bit value 0xe. Or even if the first mask uses 0b111 instead of 0b1111, a value of 0b111 would result in 0b110.

Or do we need to worry about this? Is there currently ever a valid reason to have a value > 4 in any of the `DRAW_COLORS`? Might that be something we want to use in the fugure?

site/docs/guides/basic-drawing.md
```diff
@@ -322,8 +324,13 @@ function pixel (x: i32, y: i32): void {
     const shift = u8((x & 0b11) << 1);
     const mask = u8(0b11 << shift);
 
-    // Use the first DRAW_COLOR as the pixel color
-    const color = u8(load<u16>(w4.DRAW_COLORS) & 0b11);
+    // Use the first DRAW_COLOR as the pixel color.
+    const palette_color = u8(load<u16>(w4.DRAW_COLORS) & 0b1111);
```
joshgoebel, line 328: This still allows two large a range.. you'd really want to subtract by 1 first, then mask to `0b11`.

This would allow a color of 6 for example to slip thru...

sporksmith: Subtracting 1 first could underflow. I guess that's probably ok in a lot of these, though is UB in zig.

sporksmith: You're right though that we should do something to handle values > 4. I was thinking mask again with 0b11 after the subtraction.

joshgoebel: That would have strange behavior since your'e still allowing invalid data.  I'd rather strip the invalid data entirely.

To get it 100% right it seems you first need to check for 0 right off the bat and just return (do nothing)... after that it's as simple as -1, mask.

sporksmith:
Going to wait for a maintainer to chime in how best to handle values of 5-15 before trying to edit them all again :)

sporksmith:
Went ahead with current approach + masking the final value to 2 bits. This is also what w4 does internally, e.g. here https://github.com/aduros/wasm4/blob/3bacd688eddbc1fc06a58445a74fa2bab7bf886a/runtimes/web/src/framebuffer.js#L243

PR #275, opened 2022-01-05
Replaced the `which` command with the `command -v` for locating `wasm-opt` in the Makefile templates

ZapAnton:
If `which` cannot find the `wasm-opt` file it outputs an error message, which contains a full `PATH` printed.
It creates a lot of noise for the `make` command. The `command -v` does not print anything if the file is not found.


aduros:
Thanks! I'm not seeing a `which` error message here (Ubuntu) but I can believe it's a different story on other platforms :grin:

Issue #272, opened 2022-01-05
Show blue screen of death if cart loading fails


aduros:
For example, if trying to import an unavailable function.

`safeCall` in runtime.js needs to be changed to catch WebAssembly.LinkError and possibly other error types.

PR #276, opened 2022-01-05
Added error check for LinkError in safeCall

adamcole123:
Fix for issue #272
Added check for an instance of the WebAssembly.LinkError in safeCalll, which in turn shows blue screen

### DIFF
--- a/runtimes/web/src/runtime.js
+++ b/runtimes/web/src/runtime.js
@@ -162,7 +162,8 @@ export class Runtime {
             try {
                 await fn();
             } catch (err) {
-                if (err instanceof WebAssembly.RuntimeError) {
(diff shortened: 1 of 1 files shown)


joshgoebel:
If these errors are common and a  helpful explanation can also be attached it should be.  See:

https://github.com/aduros/wasm4/blob/64b6d20d27ef053a244cd3f18c4364e8e57c3199/runtimes/web/src/runtime.js#L383

aduros:
Yes, let's add a custom error message for this, maybe looking for the word "import" (case insensitive).

We should also catch [CompileError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/CompileError) which is rarer but could happen if there's a completely corrupted wasm cart.

PR #280, opened 2022-01-05
Try to avoid calling update while we're resetting the game

DenialAdams:
This PR is totally untested, I just can't figure out how to test the site locally so i'm abusing the netlify deploy thingy

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -99,8 +99,12 @@ async function loadCartWasm () {
 
     async function reboot () {
         runtime.reset(true);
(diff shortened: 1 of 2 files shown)


DenialAdams:
> This PR is totally untested, I just can't figure out how to test the site locally so i'm abusing the netlify deploy thingy

Darn, doesn't seem like it triggered :) i'll try making it ready for review (but it's not really)

aduros:
I had disabled the netlify build trigger because we were getting close to the usage limit last month (lots of PRs!), but I just re-enabled it. It should work on future PRs :smile:

PR #283, opened 2022-01-06
Avoid calling update while we're resetting the game

### LINKED ISSUE #273: Racer on wasm4 site blue-screens on 'R'
https://wasm4.org/play/racer/

With the game running, hitting 'r' gives a wasm4 bluescreen for me on Firefox. It doesn't happen in other games on the site.

DenialAdams:
This PR fixes #273 by setting "crashed" to true after resetting the runtime. This causes the runtime not to update while the new WASM is being initialized, as per my theory in #273. Experimentally, this seems to resolve the issue.

This approach feels a little hacky, as we're semantically overloading the meaning of the "crashed" field to also mean "rebooting". Setting "paused" to true would not be as reliable, because other events on the page could still seemingly cause the runtime to unpause while it's still not ready.

A potential alternative would be to have reboot create a new runtime from scratch, avoiding any possibility of the active runtime being in an intermediate state. That may be the best approach, but I didn't implement that as it would be more refactoring and I was curious what your thoughts were @aduros

### DIFF
--- a/runtimes/web/src/constants.js
+++ b/runtimes/web/src/constants.js
@@ -274,3 +274,9 @@ export const FONT = Uint8Array.of(
 );
 
 export const showDevToolsQueryKey = 'devtools-open';
+
+export const pauseFlags = {
+    unfocused: 1,
+    crashed: 2,
+    rebooting: 4
+}
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -99,8 +99,12 @@ async function loadCartWasm () {
 
     async function reboot () {
         runtime.reset(true);
+        // By indicating that the runtime has crashed,
+        // we avoid calling update until we're done rebooting
+        runtime.crashed = true;
         await runtime.load(wasmBuffer);
         runtime.start();
+        runtime.crashed = false;
     }
 
     function takeScreenshot () {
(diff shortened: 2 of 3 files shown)


DenialAdams:
You can actually see this problem in games other than Racer too, even if they don't crash. If I reset my game "endless runner" many times, I occasionally see some weird graphics for ~1 frame. After this change (looking at the netlify preview), I can't cause that to occur

aduros:
We currently have vars `crashed` and `paused` that do basically the same thing. We could introduce another similar var called `rebooting`, but I wonder if we can combine all these pause vars into a single bitset `pauseFlags`.

`Runtime.update()` would return early if `this.pauseFlags != 0`. Pausing/resuming the runtime would involve toggling a specific bit flag, eg: `runtime.pauseFlags |= PAUSE_CRASHED`

DenialAdams:
> We currently have vars `crashed` and `paused` that do basically the same thing. We could introduce another similar var called `rebooting`, but I wonder if we can combine all these pause vars into a single bitset `pauseFlags`.
> 
> `Runtime.update()` would return early if `this.pauseFlags != 0`. Pausing/resuming the runtime would involve toggling a specific bit flag, eg: `runtime.pauseFlags |= PAUSE_CRASHED`

Nice idea! Implemented.

I'm far from a javascript dev, so sorry if the naming isn't quite right. Feel free to tweak it.

aduros:
Perfect, thanks!

Issue #265, opened 2021-12-30
Improve png2src


christopher-kleine:
The tool `w4 png2src` converts PNG files to the selected source-code.

This is useful. But it has some limitations:
1. It has to be a PNG file; other formats are not supported
2. It has to be Index; Images in RGB(A) format are not accepted

I suggest we change those two points.

Point 1 requires we taka a look at what other libs are out there.

Point 2 can be overcome if we simply create a color-map on the fly.

I'll look into this and try to get it done, before we reach v2.2.0

PR #293, opened 2022-01-09
Improved png2src

christopher-kleine:
This changes the behavior of png2src. It now ignores the palette of the image and just looks at the colors that are *actually* used.

Afterwards the colors get sorted and then applied. I tested this with the images below and it seems to work as expected.

I tried the following scenarios:

1. Image with 4 colors + Transparency (5 colors in total)
2. Image with 4 colors without Transparency (4 colors in total)
3. Image with 2 colors (2 colors in total)

The first image *should* get rejected. And that is the case.
The Image with 4 colors was tested in Index and RGB mode. Both produce the same result.
This is true for the Image with 2 colors as well.

In case of 2 colors, I tried something else to disrupt the old and the new png2src: I used 4 colors in the index, but only used 2 of those colors for the actual image. The old version detects this as a 2 Bit Image. The new version however sees this as a 1 Bit Image. This is because - like I said in the beginning - this version only cares about what is *actually* used.

// Edit: Changing the order of the colors in an indexed image doesn't effect the outcome of the processing as long as the colors itself stay the same.

// Edit 2: This should solve #265

### DIFF
--- a/cli/lib/png2src.js
+++ b/cli/lib/png2src.js
@@ -120,7 +120,8 @@ function run(sourceFile) {
         inputColorType: 1,
     });
 
-    const palette = [];
+    const palette = {};
+    let colorCount = 0;
     for (let y = 0; y < png.height; ++y) {
         for (let x = 0; x < png.width; ++x) {
             const idx = 4 * (png.width * y + x);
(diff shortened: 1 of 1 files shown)


cli/lib/png2src.js
```diff
@@ -114,76 +114,82 @@ const ALIASES = {
...
+        for (let x = 0; x < png.width; ++x) {
+            const idx = 4 * (png.width * y + x);
+            const r = png.data[idx];
+            const g = png.data[idx + 1];
+            const b = png.data[idx + 2];
+            const a = png.data[idx + 3];
+            const packed = (r << 24) | (g << 16) | (b << 8) | a;
+            let s = palette.findIndex(v => v === packed);
+            if (s === -1) {
+                palette.push(packed);
```
aduros, line 134: Can we abort early here if the palette would grow to larger than 4? We can also print the coordinate of the offending pixel.

cli/lib/png2src.js
```diff
@@ -114,76 +114,82 @@ const ALIASES = {
...
         const r = png.data[idx];
-        const g = png.data[idx+1];
-        const b = png.data[idx+2];
-        const a = png.data[idx+3];
+        const g = png.data[idx + 1];
+        const b = png.data[idx + 2];
+        const a = png.data[idx + 3];
         const packed = (r << 24) | (g << 16) | (b << 8) | a;
-        return palette.get(packed);
+        return palette.findIndex(v => v == packed);
```
aduros, line 172: Doing a `findIndex` on every pixel is kinda slow. It would be better to keep a Map of RGBA color to palette index.

cli/lib/png2src.js
```diff
@@ -114,76 +114,82 @@ const ALIASES = {
...
+            const a = png.data[idx + 3];
+            const packed = (r << 24) | (g << 16) | (b << 8) | a;
+            let s = palette.findIndex(v => v === packed);
+            if (s === -1) {
+                palette.push(packed);
+            }
+        }
     }
+    palette.sort();
+    console.log(palette);
```
aduros, line 139: Please remove debug logging :wink:

aduros (changes requested):

cli/lib/png2src.js
christopher-kleine: Whoops! Didn't notice!

cli/lib/png2src.js
christopher-kleine: While I don't think this will affect the overall performance (you don't call this 60 times per second on big images), I'll change it nonetheless.

cli/lib/png2src.js
christopher-kleine: Sure thing. I'm on it.

aduros:
Thank you very much! This is a huge improvement!

PR #294, opened 2022-01-09
libretro: State that WASM-4 requires content to run

RobLoach:
This makes it so that you after loading the WASM-4 core, the "Start Core" option isn't available until you explicitly load a game.


aduros:
Thanks!

PR #299, opened 2022-01-11
Replaced git:// urls with https://

desttinghim:
The git protocol is no longer supported by github because it is insecure https://github.blog/2021-09-01-improving-git-protocol-security-github/


aduros:
Good to know, thanks!

PR #301, opened 2022-01-11
Allow recording of videos at any duration

TLAlice:
Changed the recordVideo function so that instead of pressing record and then recording for 4 seconds, gameplay starts being recorded upon pressing record until pressing it again

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -121,8 +121,11 @@ async function loadCartWasm () {
 
     let videoRecorder = null;
     function recordVideo () {
-        if (videoRecorder != null) {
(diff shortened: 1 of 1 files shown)


aduros:
Hi Alice, thanks for the PR!

I think we need to keep some sort of maximum timeout for the video recording. Otherwise it's likely that the user will forget to press the record key a second time, which will probably run the browser out of memory and crash the tab :grin: 

Can you change it to keep the `setTimeout`, but increase it to 30-60 seconds, but still allow you to press record again to stop early?

joshgoebel:
Do we not show any indication that the video is being recored to remind them?

Issue #297, opened 2022-01-10
Switch all LLVM-based project templates to use `--stack-first`


aduros:
1. Add `--stack-first` to linker flags.
2. Increase `stack-size` to 14752 (6560 base + 8192 stack)
3. Remove `--global-base` (it's incompatible with `--stack-first`)

- [x] C / C++
- [x] D
- [x] Go
- [x] Nelua
- [x] Nim
- [x] Odin
- [x] Rust
- [x] Zig
- [ ] AssemblyScript? Not LLVM but maybe they have a similar option if it's not already default.

Refer to #255

ZetaNumbers:
@aduros `--global-base` is compatible with `--stack-first`. It's just that `--global-base` sets a base address for data segment, but not the stack.

ZetaNumbers:
Zig not placing `--stack-first` on wasm is a bug: https://github.com/ziglang/zig/issues/10567

ZetaNumbers:
AssemblyScript does not place stack first by default and there is no such option.

aduros:
> `--global-base` is compatible with `--stack-first`. It's just that `--global-base` sets a base address for data segment, but not the stack.

Have you tested this? That is how I would assume it should work, but `--global-base` seems to be ignored when `--stack-first` is used. I don't know if this is a bug or by design. It caused me a ton of confusion when originally creating the Rust project template, because Rust silently adds `--stack-first`.

PR #305, opened 2022-01-12
Switch zig template to use --stack-first

ZetaNumbers:
Implements #297 for zig template.
Require minimal zig version of 0.9.1 or 0.10.0-dev.258 to not cause silent data corruptions.


aduros:
Perfect, thanks!

PR #306, opened 2022-01-12
Fall back to old stack layout on stable zig

sporksmith:


sporksmith:
+ @ZetaNumbers

cli/assets/templates/zig/build.zig
```diff
@@ -2,41 +2,42 @@ const std = @import("std");
...
-    const zig_version = @import("builtin").zig_version;
-    if (zig_version.major > 0) return;
-    if (zig_version.minor > 10) return;
+fn version_supports_stack_first(zig_version: std.SemanticVersion) !bool {
+    if (zig_version.major > 0) return true;
+    if (zig_version.minor > 10) return true;
 
     if (zig_version.minor == 10) {
         if (zig_version.pre) |pre| {
             std.debug.print("{s}", .{pre});
```
ZetaNumbers, line 11: Oh, my dirty debug code is merged! Remove it if you like.

ZetaNumbers (approved):
Seems fine to me!

cli/assets/templates/zig/build.zig
sporksmith: Done

ZetaNumbers (approved):

sporksmith:
Verified that the template now builds and runs on both 0.9.0 (current stable) and 0.10.0-dev.258+349a7cc27

sporksmith:
@ZetaNumbers I went back and took your suggestion for simplifying the version check. Also wrote some tests to verify the expected behavior. PTAL

sporksmith:
Actually, do we know that the fix would be backported to a 0.9.1 release? If not maybe we should just check for >= 0.10.0-dev.258

sporksmith:
I don't see the fix on the 0.9.x branch. Going to add some more comments and drop the assumption that 0.9.1 will include it https://github.com/ziglang/zig/commits/0.9.x

ZetaNumbers:
> I don't see the fix on the 0.9.x branch. Going to add some more comments and drop the assumption that 0.9.1 will include it https://github.com/ziglang/zig/commits/0.9.x

https://github.com/ziglang/zig/milestone/15?closed=1

cli/assets/templates/zig/build.zig
```diff
@@ -1,42 +1,46 @@
...
-            if (number >= 258) return;
-        } else return;
-    }
+test "stack version check" {
+    const expect = std.testing.expect;
+    const parse = std.SemanticVersion.parse;
+    try expect(!try version_supports_stack_first(try parse("0.8.0")));
+
+    try expect(!try version_supports_stack_first(try parse("0.9.0")));
+    try expect(!try version_supports_stack_first(try parse("0.9.1")));
```
ZetaNumbers, line 17:
```suggestion
    try expect(!try version_supports_stack_first(try parse("0.9.1-dev.0")));
    try expect(try version_supports_stack_first(try parse("0.9.1")));
```

cli/assets/templates/zig/build.zig
```diff
@@ -1,42 +1,46 @@
...
-fn check_zig_version() !void {
-    const zig_version = @import("builtin").zig_version;
-    if (zig_version.major > 0) return;
-    if (zig_version.minor > 10) return;
+// Returns true if the version includes https://github.com/ziglang/zig/pull/10572/commits.
+// When this is false, trying to place the stack first will result in data corruption.
+fn version_supports_stack_first(zig_version: std.SemanticVersion) !bool {
+    // TODO, also allow corresponding 0.9.x+ if and when the fix has been backported.
+    // As of today it hasn't been backported yet: https://github.com/ziglang/zig/commits/0.9.x
+    return zig_version.order(try std.SemanticVersion.parse("0.10.0-dev.258")).compare(.gte);
```
ZetaNumbers, line 8:
```suggestion
    if (zig_version.pre) |pre| {
        // assert we understand the format
        try std.testing.expect(pre.len >= 5);
        try expectEqualStrings(pre[0..4], "dev.");
        try std.fmt.parseUnsigned(u32, pre[4..], 10);
        
        return zig_version.order(try std.SemanticVersion.parse("0.10.0-dev.258")).compare(.gte);
    } else return zig_version.order(try std.SemanticVersion.parse("0.9.1")).compare(.gte);
```

ZetaNumbers (changes requested):

sporksmith:
> https://github.com/ziglang/zig/milestone/15?closed=1

I see they assigned the [issue](https://github.com/ziglang/zig/issues/10567) to the 0.9.1 milestone, but I still don't see a merged fix in the 0.9.x branch.

> Also i am subscribed to zig releases, so i will check it anyway.

I'd lean towards conservatively keeping the old behavior on 0.9.x until we see the fix merged. Wdyt?

cli/assets/templates/zig/build.zig
sporksmith: We shouldn't need to parse the `pre` tag. Semantic versioning already specifies an ordering over `pre` tags, which is implemented in the `SemanticVersion` class.

sporksmith: See 11.4 https://semver.org/#spec-item-11

ZetaNumbers:
> I see they assigned the [issue](https://github.com/ziglang/zig/issues/10567) to the 0.9.1 milestone, but I still don't see a merged fix in the 0.9.x branch.

They commit to the nightly until they need 0.9.x branch, then they will bump it up to `0.9.1`. I think it would be tedious to manage one branch of an org everytime there's fix on the nightly. By your logic there's no indicaion they are working on `0.9.1` currently?

> I'd lean towards conservatively keeping the old behavior on 0.9.x until we see the fix merged. Wdyt?

We are preparing for a release too, so we should look forward too.

sporksmith:
Yeah, confirmed on https://github.com/ziglang/zig/issues/10567 that while it hasn't been cherry-picked, it's slated to be for the 0.9.1 release.

I'd still lean toward keeping the old behavior until it's actually cherry-picked, or maybe until the release is actually cut, in the case that it doesn't get cherry-picked, or gets cherry-picked and then reverted before release for whatever reason.

Being in a situation where 0.9.1 doesn't put the stack first even though it could is not so bad. Being in a situation where 0.9.1 doesn't support it but we have a release of wasm4 that tries to do it in the tutorial anyway would be pretty bad.

sporksmith:
Ah, they just pushed it. I'll go ahead and enable on 0.9.1

cli/assets/templates/zig/build.zig
ZetaNumbers: Yes, however i added assertion just to make sure there's nothing like `0.10.0-bootstrap.1`. Something like `0.10.0-alpha.1` would be less than `0.10.0-dev.258` by these rules, but that would be wrong, so i think we better handle only nighlty's `dev-*` and stable.

sporksmith: IMO them doing that would be a violation of the semver spec, but I don't know their processes well enough to be confident that they won't do it anyway. Went ahead and put in a check for "dev.".

cli/assets/templates/zig/build.zig
sporksmith: Added similar

cli/assets/templates/zig/build.zig
```diff
@@ -2,41 +2,41 @@ const std = @import("std");
...
     lib.setTarget(.{ .cpu_arch = .wasm32, .os_tag = .freestanding });
     lib.import_memory = true;
     lib.initial_memory = 65536;
     lib.max_memory = 65536;
-    lib.stack_size = 14752;
+    if (try version_supports_stack_first(zig_version)) {
+        lib.stack_size = 14752;
+    } else {
+        // `--stack-first` option have been reenabled on wasm targets with https://github.com/ziglang/zig/pull/10572
+        std.log.warn("Zig {} doesn't support stack-first. Stack overflows will result in silent data corruption.", .{zig_version});
```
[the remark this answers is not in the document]
sporksmith: Done-ish. I put in the concrete version instead of "nightly"

sporksmith:
Oops, it's wrong for 0.10.1. Pushing One More Fix (TM)

sporksmith:
Oh it was right after all; forgot I'd already covered that at the very beginning. Added a couple more test cases instead.

cli/assets/templates/zig/build.zig
```diff
@@ -14,7 +25,12 @@ test "stack version check" {
     try expect(!try version_supports_stack_first(try parse("0.8.0")));
 
     try expect(!try version_supports_stack_first(try parse("0.9.0")));
-    try expect(!try version_supports_stack_first(try parse("0.9.1")));
+    try expect(!try version_supports_stack_first(try parse("0.9.1-dev.259")));
+    try expect(try version_supports_stack_first(try parse("0.9.1")));
+
+    // Conservatively don't recognize tags other than 'dev'.
+    try expect(!try version_supports_stack_first(try parse("0.10.0-aev.259")));
+    try expect(!try version_supports_stack_first(try parse("0.10.0-zev.259")));
```
ZetaNumbers, line 33: I think we should fail on exotic zig versions, then the user would handle this code themselves to fix it. We could add a recommendation on the website to download nightly or stable 0.9.1. wdyt?

cli/assets/templates/zig/build.zig
```diff
@@ -3,9 +3,20 @@ const std = @import("std");
...
+    if (zig_version.order(try std.SemanticVersion.parse("0.10.0")).compare(.gte)) {
+        // Merged here: https://github.com/ziglang/zig/pull/10572
+        return true;
+    }
+    if (zig_version.major == 0 and zig_version.minor == 10) {
+        // Check for 0.10.0-dev.258+. Conservatively check the prefix of the tag
+        // in case zig uses other prefixes that don't respect semver ordering.
+        if (zig_version.pre) |pre| {
+            // Merged here: https://github.com/ziglang/zig/pull/10572
+            return std.mem.startsWith(u8, pre, "dev.") and zig_version.order(try std.SemanticVersion.parse("0.10.0-dev.258")).compare(.gte);
```
ZetaNumbers, line 15: Mentioned below. Everything else seems fine.

cli/assets/templates/zig/build.zig
sporksmith: I slightly prefer the current behavior of just falling back to safe legacy behavior and printing a warning. No need to force them into a session of debugging our build script to be able to build at all.

cli/assets/templates/zig/build.zig
```diff
@@ -1,42 +1,66 @@
...
     lib.setTarget(.{ .cpu_arch = .wasm32, .os_tag = .freestanding });
     lib.import_memory = true;
     lib.initial_memory = 65536;
     lib.max_memory = 65536;
-    lib.stack_size = 14752;
+    if (try version_supports_stack_first(zig_version)) {
+        lib.stack_size = 14752;
+    } else {
+        // `--stack-first` option have been reenabled on wasm targets with https://github.com/ziglang/zig/pull/10572
+        std.log.warn("Update to Zig >=0.9.1 or >=0.10.0-dev.258 to detect stack overflows at runtime.", .{});
```
ZetaNumbers, line 60:
```suggestion
        std.log.warn("Update to Zig >=0.9.1 (or >=0.10.0-dev.258 for nightly) to detect stack overflows at runtime.", .{});
```

cli/assets/templates/zig/build.zig
sporksmith: > We could add a recommendation on the website to download nightly or stable 0.9.1

Probably not worth it IMO. It's another place we'd need to keep a version number up to date. "For best results use a recent version" mostly goes without saying :)

cli/assets/templates/zig/build.zig
ZetaNumbers: I am not in the zig community, but i think `>=0.10.0-dev.258` would seem a bit too weird to a newbie.

ZetaNumbers (changes requested):

cli/assets/templates/zig/build.zig
ZetaNumbers: Well then, the descision is behind @aduros

ZetaNumbers (approved):

PR #307, opened 2022-01-12
Initialize COM for audio support on native windows

DenialAdams:
Tested on Windows 10 x64


aduros:
Awesome, thank you so much! :speaker:

PR #309, opened 2022-01-12
LIBRETRO: Add Emscripten compilation

phcoder:


ZetaNumbers:
Emscripten runs on js glue code, no? How could we add this?

phcoder:
> Emscripten runs on js glue code, no? How could we add this?

Libretro-emscripten scripts handle this already

aduros:
Platforms all the way down!

ZetaNumbers:
But does it use JS glue code?

PR #311, opened 2022-01-12
Added Player 2 Swap (F7 key) + remapped Player 2 controls

MrRafael-dev:
Press F7 to cycle between gamepads 2, 3 and 4.

Remapped Player 2 controls to numpad:
* Num 8: Up
* Num 5: Down
* Num 4: Left
* Num 6: Right
* Num 7/Num 0: X Button
* Num 9/Num 1: Z Button

### DIFF
--- a/runtimes/web/src/index.js
+++ b/runtimes/web/src/index.js
@@ -57,6 +57,7 @@ async function loadCartWasm () {
     document.getElementById("content").appendChild(canvas);
     let wasmBuffer = await loadCartWasm();
     await runtime.load(wasmBuffer);
+    let playerSwapIdx = 0;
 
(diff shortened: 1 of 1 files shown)


DenialAdams:
The new player 2 bindings aren't friendly to numpadless keyboards / laptops - something to consider

aduros:
Thanks, but let's keep the player 2 controls as they were since many keyboards don't have a numper pad. We can add numpad aliases for player 2 in another PR.

I'm going to re-implement this as a toggle because I think that's the easiest way for people to understand the intent of this feature, but we can revisit this soon. Maybe we can find space on the keyboard for player 3 and 4 :smiley:

PR #318, opened 2022-01-13
fix devtools fps counter

pfgithub:
deltaFrame was being used for the fps counter, but it is inaccurate because it does not reflect a real time. this caused the frame counter to display ~40-45fps when the actual framerate was 60fps


aduros:
Does this look good to you @FaberVitale?

FaberVitale:
> Does this look good to you @FaberVitale?

Thank you @pfgithub: LGTM! 

I think that we still need to display the average fps instead of using the latest sampled `deltaFrame` but I'm going to implement that in an other PR.

PR #319, opened 2022-01-13
Refactor blitting to fix #11

### LINKED ISSUE #11: Wrong source region when using BLIT_FLIP_X/Y with a partially clipped sprite

JerwuQu:
Also fixes rotating of non-square sprites which was I noticed was broken whilst fixing this.

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -319,18 +319,27 @@ void w4_framebufferBlit (const uint8_t* sprite, int dstX, int dstY, int width, i
 
     uint16_t colors = *drawColors;
 
+    // Clip rectangle to screen
+    int clipXMin, clipYMin, clipXMax, clipYMax;
     if (rotate) {
         flipX = !flipX;
+        clipXMin = w4_max(0, dstY) - dstY;
+        clipYMin = w4_max(0, dstX) - dstX;
+        clipXMax = w4_min(width, HEIGHT - dstY);
+        clipYMax = w4_min(height, WIDTH - dstX);
+    } else {
+        clipXMin = w4_max(0, dstX) - dstX;
+        clipYMin = w4_max(0, dstY) - dstY;
+        clipXMax = w4_min(width, WIDTH - dstX);
+        clipYMax = w4_min(height, HEIGHT - dstY);
     }
 
-    for (int y = 0; y < height; y++) {
-        for (int x = 0; x < width; x++) {
+    // Iterate pixels in rectangle
+    for (int y = clipYMin; y < clipYMax; y++) {
+        for (int x = clipXMin; x < clipXMax; x++) {
             // Calculate sprite target coords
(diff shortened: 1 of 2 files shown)


JerwuQu:
I used this code to play around with locations and blit flags
```c
#include "wasm4.h"

#define testRectWidth 16
#define testRectHeight 12
const uint8_t testRect[48] = { 0x11,0x11,0xbb,0xbb,0x55,0x55,0xaa,0xaa,0x15,0x55,0xaa,0xab,0x55,0x55,0xaa,0xaa,0x15,0x55,0xaa,0xab,0x55,0x55,0xaa,0xaa,0xff,0xff,0x00,0x00,0xbf,0xff,0x00,0x01,0xff,0xff,0x00,0x00,0xbf,0xff,0x00,0x01,0xff,0xff,0x00,0x00,0xbb,0xbb,0x11,0x11 };

void update () {
	*DRAW_COLORS = (1 << 0) | (2 << 4) | (3 << 8) | (4 << 12);

	const uint32_t flags = BLIT_FLIP_X | BLIT_FLIP_Y | BLIT_ROTATE;
	const uint32_t shift = 10;

	for (uint16_t x = 0; x < SCREEN_SIZE; x += 20) {
		for (uint16_t y = 0; y < SCREEN_SIZE; y += 20) {
			blit(testRect, x + shift, y + shift, testRectWidth, testRectHeight, BLIT_2BPP | flags);
			// blitSub(testRect, x + shift, y + shift, 10, 10, 3, 3, testRectWidth, BLIT_2BPP | flags);
		}
	}
}
```

runtimes/native/src/framebuffer.c
```diff
@@ -317,38 +317,28 @@ void w4_framebufferTextUtf16 (const uint16_t* str, int byteLength, int x, int y)
...
-            }
-            if (flipY) {
-                sy = clipYMax - sy - 1;
+    for (int y = 0; y < height; y++) {
+        for (int x = 0; x < width; x++) {
+            // Calculate sprite target coords
+            const int tx = dstX + (rotate ? y : x);
+            const int ty = dstY + (rotate ? x : y);
+            if (tx < 0 || ty < 0 || tx >= WIDTH || ty >= HEIGHT) {
+                continue;
```
aduros, line 332: Instead of doing a clipping check on every pixel, I wonder if it would it be possible to keep it as before where we precalculate the clipped rectangle?

aduros (changes requested):
Thanks a lot for invesigating this! It will be awesome to have that old bug finally fixed :smiley:

JerwuQu:
Since the code now also accounts for rotating non-square sprites, the size calculation becomes a bit messier, which is why I initially went with the other solution (which I think is easier to understand at a glance).

aduros:
Looks great, thanks!

PR #322, opened 2022-01-15
png2src: Readd support for indexed PNG ordering

JerwuQu:
I was scratching my head a bit at why I couldn't manually order the palette in a png and then use png2src (since I could do so before!) After looking at the code, I noticed that functionality had been lost in #293.

This PR changes so when a png has an indexed palette, it will use that order, otherwise it will proceed to use the algorithm implemented in #293.

### DIFF
--- a/cli/lib/png2src.js
+++ b/cli/lib/png2src.js
@@ -120,37 +120,38 @@ function run(sourceFile) {
         inputColorType: 1,
     });
 
-    const palette = {};
+    const palette = new Map();
     let colorCount = 0;
-    for (let y = 0; y < png.height; ++y) {
-        for (let x = 0; x < png.width; ++x) {
-            const idx = 4 * (png.width * y + x);
-            const r = png.data[idx];
-            const g = png.data[idx + 1];
-            const b = png.data[idx + 2];
-            const a = png.data[idx + 3];
-            const packed = (r << 24) | (g << 16) | (b << 8) | a;
-            let s = palette[packed]
-            if (!s) {
-                if (colorCount >= 4) {
-                    let RGBs = [];
-                    for (let key in palette) {
-                        RGBs.push(`- (R: ${palette[key].r}, G: ${palette[key].g}, B: ${palette[key].b}, A: ${palette[key].a}); first seen at (${palette[key].x}, ${palette[key].y})`)
+    if (png.palette) {
+        colorCount = png.palette.length;
+        if (colorCount >= 4) {
+            throw new Error('Indexed PNG has too many colors: maximum is 4.');
+        }
+        for (let i = 0; i < colorCount; ++i) {
+            const rgba = png.palette[i];
+            const packed = (rgba[0] << 24) | (rgba[1] << 16) | (rgba[2] << 8) | rgba[3];
+            palette.set(packed, i);
+        }
+    } else {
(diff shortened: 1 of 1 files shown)


christopher-kleine:
Hi!

The current behavior is actually intended. I couldn't find it in the PR, so it must have been in Discord I guess.
@aduros mentioned that the behavior would change with #293 , but that was fine. The idea is to use only the colors you *actually* use.
The indexed approach came with another problem (for me at least). If you have an image that *used* 2 colors, but had a palette with 4 colors, it was invalid from the start. Or even a valid Image: You used 2 colors, palette was 4 colors: The old png2src wasted a lot of space in this scenario, since it used the 2 Bit Version instead of the more optimal 1 Bit Version.

Aside from that: Your PR doesn't work. Or to more specific: 2 Bit Indexed images don't.
I have a small set of test images. Each are 16x16 Pixel. And the following cases are preset.

Here is an overview of the test result:

| Case                 | Old     | Current | This PR |
|----------------------|---------|---------|---------|
| Indexed (5 Colors)   | Invalid | Invalid | Invalid |
| RGB (5 Colors)       | Invalid | Invalid | Invalid |
| Indexed (2/2 Colors) | 1 Bit   | 1 Bit   | 1 Bit   |
| Indexed (2/4 Colors) | 2 Bit   | 1 Bit   | Invalid |
| Indexed (2/5 Colors) | Invalid | 1 Bit   | Invalid |
| RGB (2 Colors)       | Invalid | 1 Bit   | 1 Bit   |
| Indexed (4/4 Colors) | 2 Bit   | 2 Bit   | Invalid |
| Indexed (4/5 Colors) | Invalid | 2 Bit   | Invalid |
| RGB (4 Colors)       | Invalid | 2 Bit   | 2 Bit   |

cli/lib/png2src.js
```diff
@@ -120,37 +120,38 @@ function run(sourceFile) {
...
-            if (!s) {
-                if (colorCount >= 4) {
-                    let RGBs = [];
-                    for (let key in palette) {
-                        RGBs.push(`- (R: ${palette[key].r}, G: ${palette[key].g}, B: ${palette[key].b}, A: ${palette[key].a}); first seen at (${palette[key].x}, ${palette[key].y})`)
+    if (png.palette) {
+        colorCount = png.palette.length;
+        if (colorCount >= 4) {
+            throw new Error('Indexed PNG has too many colors: maximum is 4.');
+        }
```
christopher-kleine, line 129: This refuses 4 color palette indexed images.

JerwuQu: Good catch

JerwuQu:
> The idea is to use only the colors you _actually_ use.

This is a fair point.

I just found this to be the least surprising option. It's a 1:1 map and will tell you when it doesn't work. Dropping unused colors as you say is fine, but loses the ability to freely move around 3 colors spread within a 2BPP palette (unless that use case is also considered specifically for 4 color indexed pngs). 

Currently there's no way to order the palette, but I noticed there used to be a `.sort()` when it was an array. Ordering by value as you were discussing in Discord is a good method if the "3 colors spread within a 2BPP palette" use case is deemed too niche.

Can close this in favor of readding the `.sort()` instead. I don't care too much which way it goes as long as there is *some* way to do it.

Issue #320, opened 2022-01-14
Add panning to `tone`


JerwuQu:
Two new flags `TONE_PAN_LEFT` and `TONE_PAN_RIGHT` to pan a sound completely left or completely right, similar to on a Gameboy.

aduros:
I like that the proposed implementation is elegant and a non-breaking change, but I'm reluctant to add new functionality at the moment. I feel like the use-cases for stereo sound in 2D games are kind of limited, and especially in 2D retro games, so not sure if it's worth the additional API surface.

joshgoebel:
In my quick googling it seems stereo was pretty common on Gameboy at least (if you plugged in headphones):

> All the Pokemon games (except for the original red and blue) support stereo sound but it needs to be enabled in the options menu. It makes for a pretty fun headphone experience. The soundtracks are great in stereo, and in battles, the sounds come from either side of the screen.

 Stereo seems an easy (even at the time) way to add extra depth of experience to 2D games of the time - so it's not a surprise people took advantage of that. Not sure about other retro systems.  I played so much on TV and doubtful I would have noticed the stereo.

> I feel like the use-cases for stereo sound in 2D games are kind of limited,

I mean if we just simply don't want more features (or API surface) that's your call, but I think I'd have to disagree with "use cases are limited" being the justification.  Just my 0.02.

JerwuQu:
I don't think it's limited in usage at all when considering music, but it can also make for cooler sound effects. I see it as a small addition that will have big impact.

As @joshgoebel said, it was pretty common and a lot of games used it.
Here's a couple of examples: [Final Fantasy Adventure: Endless Battle](https://www.youtube.com/watch?v=Or0qv44SVmo), [Super Mario Land's Overworld Theme](https://www.youtube.com/watch?v=Gb33Qnbw520).

sporksmith:
Another example I have fond memories of on gameboy, Final Fantasy Legend: https://youtu.be/BoeKZq6vNfQ

PR #330, opened 2022-01-16
Tone panning (solves #320)

JerwuQu:
🎈 

Needs:
- [X] Porting to web APU
- [X] Documentation
- [X] New slider in Sound Test Tool
- [X] Test ScriptProcessor
- [ ] Test Worklet

### DIFF
--- a/runtimes/native/src/apu.c
+++ b/runtimes/native/src/apu.c
@@ -129,7 +129,7 @@ void w4_apuTone (int frequency, int duration, int volume, int flags) {
 
     int channelIdx = flags & 0x03;
     int mode = (flags >> 2) & 0x3;
-    const int pan = ((flags >> 4) & 0x3) % 3;
+    int pan = (flags >> 4) & 0x3;
 
     // TODO(2022-01-08): Thread safety
     Channel* channel = &channels[channelIdx];
--- a/runtimes/native/src/apu.c
+++ b/runtimes/native/src/apu.c
@@ -38,6 +38,9 @@ typedef struct {
     /** Used for time tracking. */
     float phase;
 
+    /** Tone panning. 0 = center, 1 = only left, 2 = only right. */
+    uint8_t pan;
+
     union {
         struct {
             /** Duty cycle for pulse channels. */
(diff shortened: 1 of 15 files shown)


joshgoebel:
I'm not sure "panning" is the right name though if we aren't actually panning from left to right...  and it's just "which channel to play this on".

JerwuQu:
I think panning is the most commonly used term for this, despite it only being hard left, or hard right.

Looking at different resources for the Gameboy as example, they also uses "panning" as the term.

Examples:
- [GbdevWiki](https://gbdev.gg8.se/wiki/articles/Sound_Controller)
- [LSDj Manual](https://www.littlesounddj.com/lsd/latest/documentation/LSDj_9_2_6.pdf)
- [Game Boy Sound Emulation](https://nightshade256.github.io/2021/03/27/gb-sound-emulation.html)

runtimes/web/src/apu-worklet.ts
```diff
@@ -126,6 +129,7 @@ class APUProcessor extends AudioWorkletProcessor {
 
         const channelIdx = flags & 0x3;
         const mode = (flags >> 2) & 0x3;
+        const pan = ((flags >> 4) & 0x3) % 3;
```
aduros, line 132: Is this the same as:

```suggestion
        const pan = (flags >> 4) & 0x3;
```

runtimes/native/src/apu.c
```diff
@@ -126,6 +129,7 @@ void w4_apuTone (int frequency, int duration, int volume, int flags) {
 
     int channelIdx = flags & 0x03;
     int mode = (flags >> 2) & 0x3;
+    const int pan = ((flags >> 4) & 0x3) % 3;
```
aduros, line 132:
```suggestion
    int pan = (flags >> 4) & 0x3;
```

Re: const, I should probably do a style pass on the entire code base, but for now let's a least be consistent :grin:

aduros (approved):
I haven't tried it yet, but looks great! I like that setting both PAN_LEFT and PAN_RIGHT is the same as no panning.

PR #332, opened 2022-01-16
[feat][devtools] show cartridge size

FaberVitale:
### Description

This PR adds a section that displays the current cartridge size and the max size allowed `0xffff`.
It also includes a warning emoji that is displayed if wasm buffer length is above limit. 

###  Below max size

Screenshot 2022-01-16 at 19 46 28

###  Above max size

Screenshot 2022-01-16 at 19 48 53

### DIFF
--- a/devtools/web/index.html
+++ b/devtools/web/index.html
@@ -28,6 +28,7 @@
           },
           fps: 60,
           storedValue: null,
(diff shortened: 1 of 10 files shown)


aduros:
Neat, thanks!

Issue #333, opened 2022-01-16
Audio Frequency Proposal


marler8997:
Currently the WASM audio API controls the frequency of its channels using a 16-bit integer.  However, the affect audio frequencies have on the human ear is on a "logarithmic" frequency scale rather than a linear one.  By distributing the frequency values linearly, the lower frequencies will have exponentially less granularity than the higher ones.  For example, frequencies 1 and 2 are single "octave" apart, but so are 32,000 and 64,000 which would have 32,000 more pitches available in between them when using whole-number frequencies.

## Some Context about Audio frequencies

The human ear associates a doubling of the frequency with the sound of an "octave".  To humans, two frequencies an "octave" apart sound like the "same note" harmonically but at lower/higher registers.  The octave is a special interval because you can think of it like a loop.  As you increase the frequency, once you've hit an octave the same notes repeat themselves in a higher register.

This is where the logarithmic scale comes from.  You have to double the initial frequency every time to get to the next octave.  So if you start at frequency 100, the next octave would be 200, then the next would be 400 and so on.  Another way to look at it, adding 50 Hz to the frequency 100 has a greater affect than adding it to 200, you would have to add 100 to get the same affect at that frequency.

For example, the note "c0" in an "equal temperament" A440 tuning is defined as the frequency 16.35.  The next octave above that, "c1" is at 32.70 (note that it's not exactly double because of "equal temperament", another topic I won't go into here).  And the next octave "c2" is 65.41.  The interval between c0 and c1 sounds the same as the interval between c1 and c2, even though the frequency difference between each one differs by a factor of 2.  Adding 16 Hz to c0 brings us to the next octave c1, but adding 16 Hz to c1 only gets us halfway there bring us to g1 (an interval called a "fifth").

## The proposal

I propose that to set the frequency of each channel, we leverage the standard established by the MIDI protocol.

https://www.cs.cmu.edu/~music/cmsip/readings/MIDI%20tutorial%20for%20programmers.html

Assuming a standard A440 equal-temperament tuning, the frequency of any midi note can be calculated with:
```js
const twelvth_root_of_2 = Math.pow(2, 1/12);
function midiNoteToFrequency(pitch) {
    return 440 * Math.pow(twelvth_root_of_2, pitch - 69);
}
```
[...]

PR #334, opened 2022-01-17
An implementation of a new frequency mode, see issue #333

marler8997:
An implementation of the "Audio Frequency Proposal" found here: #333 

At this point the proposal has not been accepted, but I've implemented it here to demonstrate what it would take to support the proposal along with keeping the current mechanism (whole-number frequencies).

To summarize what this does, this leverages what the MIDI protocol implements for frequencies.  It uses 7-bits for the "pitch", and then the upper 8-bits can "bend" the note upwards to gain access to frequencies in between the standard 12-note musical scale.

Here's an example of how to use it:
```c
// current interface
tone(440, ...);


// new interface
*SYSTEM_FLAGS |= SYSTEM_MIDI_FREQUENCY_MODE; // this enables the new MIDI frequency mode

tone(69, ...);

// to pitch the frequency up slightly (in between pitch 69 and 70)
tone(69 | (160 << 8), ...)
```


joshgoebel:
Feels like we need a `note` API... the higher-level libs I'm writing for TIC-80 have a note API for playing notes that does the translations from notes into frequencies and the lower-level APIs...  69 isn't a "frequency" or "pitch" for A4, just a ID... so the naming is already a bit "off"...

PR #337, opened 2022-01-18
Port sample-based APU to web

### LINKED ISSUE #321: APU parity
Currently, the APU for web is implemented via methods in the Web Audio API. Aside from this causing slight differences between Chromium and Firefox, it also causes differences between native and web.

I think rewriting the web APU to be sample buffer-based like the native one would be a good idea. It would ensure that audio sounds exactly the same, and would make porting features between the two implementations much easier. I also suspect a sample-based implementation would make solving #308 easier.

Thoughts?

aduros:
Some notes:

- Noise sounds richer, especially in lower frequencies
- High frequency pulses sound a bit more harsh, since it's now a true pulse and not an approximation using sine waves
- Volume seems higher across the board, might need to tweak that
- `BUFFER_SIZE` may need to be increased, at the cost of additional latency

Closes #321

### DIFF
--- a/runtimes/native/src/apu.c
+++ b/runtimes/native/src/apu.c
@@ -86,6 +86,18 @@ static int16_t getCurrentVolume (const Channel* channel) {
     }
 }
 
+static float polyblep (float phase, float phaseInc) {
+    if (phase < phaseInc) {
+        float t = phase / phaseInc;
+        return t+t - t*t;
+    } else if (phase > 1.f - phaseInc) {
+        float t = (phase - (1.f - phaseInc)) / phaseInc;
+        return 1.f - (t+t - t*t);
+    } else {
+        return 1.f;
+    }
+}
+
 void w4_apuInit () {
     channels[3].noise.seed = 0x0001;
 }
@@ -165,8 +177,10 @@ void w4_apuWriteSamples (int16_t* output, unsigned long frames) {
                     sample = volume * channel->noise.lastRandom;
 
                 } else {
-                    channel->phase += (float)freq / SAMPLE_RATE;
-                    if (channel->phase > 1) {
+                    float phaseInc = (float)freq / SAMPLE_RATE;
+                    channel->phase += phaseInc;
+
+                    if (channel->phase >= 1) {
                         channel->phase--;
                     }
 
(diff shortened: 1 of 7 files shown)


JerwuQu:
I've been messing around on your branch a bit, and with the native version (since everything except the `BUFFER_SIZE` thing applies there too).

The biggest issue right now is that the pulse channel is aliasing (quite a lot). There are many solutions to solve this, but I think the most practical and performant would be something like [PolyBLEP](http://www.martin-finke.de/blog/articles/audio-plugins-018-polyblep-oscillator/). Other solutions involve a implementing filters or using additive synthesis instead.

About the noise channel, I'm not quite sure yet. It sounds pretty good up to ~300 Hz like you say, but once you go above and try something like 650 Hz it sounds metallically distorted and not pleasant at. This might also be a result of aliasing, but is probably easily solved using interpolation.

aduros:
Thanks a lot for the feedback and research! Later today I'll try out your proposed noise generator change, and look more into PolyBLEP.

Pinging some other folks who've been interested in audio lately: @micaevski @marler8997 @rohlem @OliveIsAWord

aduros:
@JerwuQu I used your polyblep implementation, seems to work much better than the experiments I tried!

Some final considerations before this can be merged:
- Volume needs to be adjusted. Pulse channels are louder than before, and the triangle channel is quieter than before.
- Should we bump `BUFFER_SIZE` on web? The additional latency going to 1024 is noticeable to me, so I'd rather not unless necessary.
- I'm adding a [short release duration](https://github.com/aduros/wasm4/blob/issue321/runtimes/native/src/apu.c#L148) to triangles to prevent "popping". Does this look right? I think this duration may need to be increased a bit more, as I'm still hearing some popping when the triangle channel ends.

JerwuQu:
# Volumes

My guess is that the pulses are louder than triangles simply because of how much harmonic space they take up.

Just threw a square and triangle at the same amplitude into FL Studio to demonstrate:
image
image

Harshly bitcrushing the triangle yields something like this, showing that a lower fidelity triangle would take up more soundspace while at the same time sounding more like an old console
image

# BUFFER_SIZE

It would be great if we could detect a buffer underrun and double the buffer size if it happens. That way we could keep the low BUFFER_SIZE for devices that can handle it. Disadvantage is that it could be weird if you're playing something like wasm4nia and suddenly you have double the audio lag in the middle of a song! Perhaps it would be best to detect buffer underruns within the first few seconds of gameplay and then never touch the buffer size again? I'm not sure...

# Triangle Release

Yeah, seems like an acceptable solution to me. I don't think you should be hearing popping from this since it has 44 samples to ramp down. It might be related to something else, possibly the issue with off-by-ones for ADSR timings.

aduros:
Volumes: That's interesting. I can try rounding the triangle output to 8 bits and see what happens :smiley: 

BUFFER_SIZE: Hmm, for the sake of performance maybe we should bite the bullet and implement AudioWorkletNode. I'll look into some webpack shenanigans to see if don't need a separate .js file for the worker, which wouldn't be possible in carts bundled to html. We could have a SharedArrayBuffer for just the Channel struct data.

Triangle release: I thought the off-by-one was just for the attack phase, but I'll look more into it.

Thanks @JerwuQu !

JerwuQu:
LGTM

Subjective nitpick would be to switch `>>> 0` for `| 0`, since it's more commonly used AFAIK

PR #341, opened 2022-01-18
Add support for saving/loading disks to native runtime

JerwuQu:
Will use the `<filepath>.disk` as filename for the save disk.

When running wasm files directly, the name will become `my_game.wasm.disk`.
When running bundles, the name will become `my_game_bundle.disk` (or `my_game_bundle.exe.disk` on Windows).
This could be changed to trim the extension, but since linux doesn't use file extensions for executables, the code would vary depending on OS.

The reason for exposing the Disk struct and not simply copying the struct (like libretro does) is to make disk files not dependent on endianness (and as such transferrable between different platforms).
I otherwise kept the interface for `w4_runtimeInit` as to not break libretro main (or any others).

### DIFF
--- a/runtimes/native/src/backend/main.c
+++ b/runtimes/native/src/backend/main.c
@@ -105,6 +105,18 @@ static void saveDiskFile (const w4_Disk* disk, const char *diskPath) {
     }
 }
 
+static void trimFileExtension (char *path) {
+    size_t len = strlen(path);
+    while (len--) {
+        if (path[len] == '.') {
+            path[len] = 0; // Set null terminator
+            return;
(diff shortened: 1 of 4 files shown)


aduros (changes requested):
This looks perfect, thanks!

Can we change `w4_runtimeInit` to accept a `w4_Disk*`? As long as we change both backends I don't think it will cause any breakage.

aduros:
Also maybe we could put the `.exe` filename stripping behind an `#ifdef _WIN32`?

aduros:
Thanks!

PR #343, opened 2022-01-19
Fix several big-endian bugs

phcoder:


aduros:
Thanks!

I'd like to refactor this later a bit so that the core runtime doesn't depend on wasm3. I also think we need a better strategy to handle endian bugs from popping up in the future.

PR #344, opened 2022-01-19
Define nimNoQuit

tjpalmer:
This makes use of https://github.com/nim-lang/Nim/pull/19419 but should be ignored without the Nim changes.


aduros:
Thanks Tom!

PR #345, opened 2022-01-20
Directly convert FRAMEBUFFER pointer in Rust

majaha:
Change the Rust examples in Docs > Guides > Basic Drawing to directly convert FRAMEBUFFER from a *mut to a &mut without null-checking it.


aduros:
Thanks!

@FaberVitale please let us know if this is wrong somehow.

PR #346, opened 2022-01-20
Fix off-by-one for APU timings

JerwuQu:
I messed around with APU on native and after hearing pops, noticed that the timings for `getCurrentVolume` were off by one.

As example, if you were to play a `tone` with `a`, `d`, and `r` all set to `0`, `channel->sustainTime` would be set to `time` ([see lines 106-109](https://github.com/JerwuQu/wasm4/blob/6790d3c77805eb2140fc903fa5be5e9fd2402a15/runtimes/native/src/apu.c#L106L109)), meaning it should start sustaining immediately, but since the comparator is `>`, you instead get one tick of attack, meaning you would get pops if you were to chain `tone`s after each other.

### DIFF
--- a/runtimes/native/src/apu.c
+++ b/runtimes/native/src/apu.c
@@ -72,11 +72,11 @@ static uint16_t getCurrentFrequency (const Channel* channel) {
 }
 
 static int16_t getCurrentVolume (const Channel* channel) {
-    if (time > channel->sustainTime) {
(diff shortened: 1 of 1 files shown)


aduros:
Ah, you'e right!

I'm going to leave this open for now so I don't forget to apply it to the new APU branch after it's been merged.

Issue #347, opened 2022-01-20
Ovals across screen edge wrap or something


tjpalmer:
I don't have a screenshot handy, but this at least starts tracking the issue.

Tinchetsu:
I was going to report this as well. Attaching gif from a game I'm making for wasm4 gamejam.

The bullets are ovals, when they go offscreen, they wrap.
![game](https://user-images.githubusercontent.com/10901731/150585112-4b41327f-115d-4b78-aa5b-ba61e103a900.gif)

JerwuQu:
I'll start tackling this one now

PR #352, opened 2022-01-22
Fix oval drawing bug on the web runtime

JerwuQu:
Solves issue #347

I noticed the issue only applied to web, and the reason was that the wrong line drawing method was being used.
Compare [native](https://github.com/aduros/wasm4/blob/5fc0583b5c0cd1e3b61a915f72c985f625da53b5/runtimes/native/src/framebuffer.c#L190L191) with [web](https://github.com/aduros/wasm4/blob/5fc0583b5c0cd1e3b61a915f72c985f625da53b5/runtimes/web/src/framebuffer.js#L185L186).


aduros:
Nice catch, thanks!

Issue #331, opened 2022-01-16
ADSR specification 


micaevski:
Hey there!
I just started playing with WASM-4 yesterday and I'm really enjoying it! Thank you all who are involved. 

I am building a step sequencer and I realised that the ADSR parameters of the tone function are not what you would call "Standard". 

Maybe it's some vintage specification, I don't know but usually you would have a them specified as 
Attack - time to reach the set volume level
Decay -  time to reach the sustain volume level
Sustain -  a secondary volume level that is sustained through the duration of the note. 
Release - time after the release of the note until the volume has reached 0.
This of course relies on the note having a duration. 

The current specification leads to all sorts of weird states like having volume when the volume is all the way down or really big jumps in volume just because you introduced some attack.
Also there is no way to add attack and decay but still be able to mix your tone at a lower level than maximum.

It would be really great if the `tone` function had a duration parameter so the specification could change. 

something like
`tone(freq1, envelope, duration | (volume << 16), flags)`

or even better 

`tone(freq, ampAttack | (ampDecay << 16), ampSustain | (ampRelease << 16), freqAttack | (freqDecay << 16), freqSustain | (freqRelease << 16), duration | (volume << 16), flags)` 

The envelope parameters are split so the sustain part of the frequency envelope will now have 16bits, enough to describe the sustained frequency. 
The amp sustain could be a full independent volume level or something relative to the volume parameter.

I think this allows more musical variety and is much more resembling of how synthesisers work. 

Thanks again.

JerwuQu:
I noticed this too. As far as I understood from my testing, how it works right now is that `volume` parameter is what would be Sustain in ADSR, and that a tone amplitude parameter doesn't exist, it's instead locked to 100%, making it impossible to set amplitude if you also want to use ADSR.

It would be a breaking change, but I definitely think it's worth considering making it more in line with the norm, since right now it's a bit weird.

I don't think we need 16 bits of precision for ADSR like in your examples though, and mixing duration and volume into the same bitfield might be a bit strange (since they're not related).
I personally like `tone(freq, duration, volume, adsr, flags)` better since it's easier to use for people that don't need the advanced features.
It would be used as `tone(freq1 | (freq2<< 16), duration, volume, (a << 24) | (d << 16) | (s << 8) | r), flags)`.
If following ADSR, only the `r` length would actually get added onto the total duration. Right now all of them are added together.

[...]

PR #353, opened 2022-01-22
Add peak amplitude to ADSR specification

micaevski:
This adressess issue #331 
The specification is still not standard but this solution is backwards compatible.
`tone`'s volume parameter now accepts a peak volume with an 8 bit shift.
This volume is the peak volume reached at the end of the attack duration.

### DIFF
--- a/examples/sound-demo/src/main.c
+++ b/examples/sound-demo/src/main.c
@@ -23,6 +23,7 @@ int values[] = {
     /* 255, */
     0,
     100,
+    100,
     0,
     0,
 };
@@ -34,6 +35,7 @@ const int MAX_VALUES[] = {
     255,
     255,
     255,
+    100,
     100,
     3,
     3,
@@ -91,9 +93,10 @@ void update () {
     drawControl("DECAY", x, y + 3*spacing, 3);
     drawControl("SUSTN", x, y + 4*spacing, 4);
     drawControl("RLEAS", x, y + 5*spacing, 5);
-    drawControl("VOLUM", x, y + 6*spacing, 6);
-    drawControl("CHNNL", x, y + 7*spacing, 7);
-    drawControl(" MODE", x, y + 8*spacing, 8);
+    drawControl(" PEAK", x, y + 6*spacing, 6);
+    drawControl("VOLUM", x, y + 7*spacing, 7);
+    drawControl("CHNNL", x, y + 8*spacing, 8);
+    drawControl(" MODE", x, y + 9*spacing, 9);
 
     *DRAW_COLORS = 4;
     text("Arrows: Adjust\nX: Play tone", x, 8);
@@ -105,7 +108,7 @@ void update () {
     char pressedThisFrame = gamepad & (gamepad ^ lastGamepadState);
     lastGamepadState = gamepad;
 
-    if ((pressedThisFrame & BUTTON_DOWN) && arrowIdx < 8) {
+    if ((pressedThisFrame & BUTTON_DOWN) && arrowIdx < 9) {
         ++arrowIdx;
     }
     if ((pressedThisFrame & BUTTON_UP) && arrowIdx > 0) {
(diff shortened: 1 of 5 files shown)


runtimes/native/src/apu.c
```diff
@@ -96,6 +99,9 @@ void w4_apuTone (int frequency, int duration, int volume, int flags) {
     int decay = (duration >> 16) & 0xff;
     int attack = (duration >> 24) & 0xff;
 
+    int sustainVolume = volume & 0xffff;
+    int peakVolume = (volume >> 16) & 0xffff;
```
JerwuQu, line 103: This should probably be set to 100 by default or it will change behavior of current carts

site/docs/guides/audio.md
```diff
@@ -196,16 +196,25 @@ w4.tone(262 | (523 << 16), 60, 100, w4.TONE_PULSE1);
...
+| Sustain | 0     | The volume sustained for the `sustain` phase duration. |
+| Peak    | 16    | The volume reached by the `attack` phase               |
 
 [ADSR](https://en.wikipedia.org/wiki/ADSR_envelope) describes how the volume changes over time, and
 has 4 time components:
 
-| Time        | Shift | Description                                                                  |
+| Time        | Shift | Description                                                                    |
 | ---         | ---   | ---                                                                          |
 | **A**ttack  | 24    | The time it takes to initially ramp up from 0 volume to 100% volume.         |
```
JerwuQu, line 214: Should say "peak volume" here, not 100%

JerwuQu (commented):
Haven't tried the code yet but spotted these

site/docs/guides/audio.md
micaevski: I thought I changed that 🤷

micaevski: Ahh I accidentally discarded it when removing some formatting changes.

runtimes/native/src/apu.c
micaevski: Not sure how to implement this TBH. If the peak volume input is 0. how can I tell if it's the user's input or no input?

JerwuQu: Having it set to 0 would mean that the attack phase is just silent and that the decay phase essentially becomes an attack phase. You don't gain anything from being able to do this.

micaevski: Well, I guess the way to look at it is, it just behaves the same as before but if you want you can set a peak volume as well. From that perspective it makes sense, but from a fresh user's perspective, it's weird that setting the peak to 0 results in full peak volume. This should be reflected in the docs.

JerwuQu: Yea, just say it will default to 100%. Not really that different from pitch bend not accepting a 0 either.

examples/sound-demo/src/main.c
```diff
@@ -92,8 +94,9 @@ void update () {
     drawControl("SUSTN", x, y + 4*spacing, 4);
     drawControl("RLEAS", x, y + 5*spacing, 5);
     drawControl("VOLUM", x, y + 6*spacing, 6);
-    drawControl("CHNNL", x, y + 7*spacing, 7);
-    drawControl(" MODE", x, y + 8*spacing, 8);
+    drawControl("PEAK", x, y + 7*spacing, 7);
```
aduros, line 97:
```suggestion
    drawControl(" PEAK", x, y + 7*spacing, 7);
```

runtimes/native/src/apu.c
```diff
@@ -28,7 +28,10 @@ typedef struct {
     unsigned long long releaseTime;
 
     /** Sustain volume level. */
-    int16_t volume;
+    int16_t sustainVolume;
+
+     /** Peal volume level (at the end of the attack phase). */
```
aduros, line 33:
```suggestion
     /** Peak volume level (at the end of the attack phase). */
```

aduros (changes requested):
Hi @micaevski, thanks a lot for the PR!

What do you think about fitting the sustain and peak volumes in 8 bits instead of 16? This would give us more room to use the remaining 16 bits in the future for something.

Also please add a special case where a 0% peak volume translates to 100% to keep backwards compat (and is a good default anyways)

micaevski:
@aduros @JerwuQu I made the requested changes.

aduros:
Thank you very much!

PR #360, opened 2022-01-27
Add the -Wconversion -Wsign-conversion flags to the Makefile template

ZapAnton:
Creating games using wasm4 assumes a lot of number operations.
The `-Wconversion` and `-Wsign-conversion` are added in order
to catch any subtle bugs like number overflows or casting.


aduros:
Thanks!

PR #362, opened 2022-01-27
Change Rust template static muts to consts

majaha:
This patch changes the FRAMEBUFFER and PALETTE variables in the Rust template from `pub static mut` to `pub const`, as the pointers themselves don't ever need to be mutated or have a static memory address.

In the future, we may be able to have the framebuffer directly like this:
`pub static mut FRAMEBUFFER: [u8; 6400] = unsafe { *(0xa0 as *mut [u8; 6400]) };`
but this isn't stablised yet.


aduros:
Thanks!

PR #365, opened 2022-01-28
Add START_TIME to retrieve when the runtime started

RobLoach:
Introduces a `START_TIME` constant which is an unsigned 64 bit integer representing the unix timestamp of when the runtime started. This would be helpful for creating time-sensative applciations, like clocks, calendars, or persistent games like [Tamagotchi](https://en.wikipedia.org/wiki/Tamagotchi).

This isn't ready to be merged as it...

- [ ] Needs more testing
- [ ] Needs some tweaks in some of the languages (DataView doesn't support uint64 right now for instance)
- [ ] Unsure about the name

### DIFF
--- a/devtools/web/src/constants.ts
+++ b/devtools/web/src/constants.ts
@@ -17,6 +17,7 @@ export const ADDR_MOUSE_X = 0x1a;
 export const ADDR_MOUSE_Y = 0x1c;
 export const ADDR_MOUSE_BUTTONS = 0x1e;
 export const ADDR_SYSTEM_FLAGS = 0x1f;
+export const ADDR_START_TIME = 0x20;
 export const ADDR_FRAMEBUFFER = 0xa0;
 
 export const BUTTON_X = 1;
@@ -82,6 +83,10 @@ export const memoryMap: Readonly<Record<string, Range>> = {
   SYSTEM_FLAGS: {
     offset: ADDR_SYSTEM_FLAGS,
     len: 1,
+  },
+  START_TIME: {
+    offset: ADDR_START_TIME,
+    len: 8,
   },
   RESERVED: {
     offset: 0x0020,
--- a/devtools/web/src/models/MemoryView.ts
+++ b/devtools/web/src/models/MemoryView.ts
@@ -44,6 +44,7 @@ export class MemoryView implements MemoryViewComputedProperties {
   readonly pointerPos: Point;
   readonly mouseBtnByte: number;
   readonly systemFlags: number;
+  readonly startTime: number;
   readonly gamepads: number[];
   readonly palette: [number, number, number, number];
   readonly drawColors: number;
@@ -58,6 +59,7 @@ export class MemoryView implements MemoryViewComputedProperties {
     this.palette = extractPalette(dataView);
     this.gamepads = extractGamepads(dataView, bufferedData);
     this.systemFlags = dataView.getUint8(constants.ADDR_SYSTEM_FLAGS);
+    this.startTime = dataView.getUint32(constants.ADDR_START_TIME); // TODO: DataView doesn't support uint64?
     this.mouseBtnByte =
       dataView.getUint8(constants.ADDR_MOUSE_BUTTONS) |
       bufferedData.mouseButtons;
--- a/runtimes/native/src/runtime.c
+++ b/runtimes/native/src/runtime.c
@@ -2,6 +2,7 @@
 
 #include <stdio.h>
 #include <string.h>
+#include <time.h>
 #include <m3_env.h>
 
 #include "apu.h"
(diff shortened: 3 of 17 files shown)


devtools/web/src/constants.ts
```diff
@@ -83,6 +84,10 @@ export const memoryMap: Readonly<Record<string, Range>> = {
     offset: ADDR_SYSTEM_FLAGS,
     len: 1,
   },
+  START_TIME: {
+    offset: ADDR_START_TIME,
+    len: 8,
+  },
   RESERVED: {
     offset: 0x0020,
     len: 128,
```
RobLoach, line 93:
``` suggestion
    len: 120,
```

devtools/web/src/models/MemoryView.ts
```diff
@@ -58,6 +59,7 @@ export class MemoryView implements MemoryViewComputedProperties {
     this.palette = extractPalette(dataView);
     this.gamepads = extractGamepads(dataView, bufferedData);
     this.systemFlags = dataView.getUint8(constants.ADDR_SYSTEM_FLAGS);
+    this.startTime = dataView.getUint32(constants.ADDR_START_TIME); // TODO: DataView doesn't support uint64?
```
RobLoach, line 62: `DataView` doesn't have a `getUint64()` function right now.

runtimes/web/src/runtime.js
```diff
@@ -50,6 +50,10 @@ export class Runtime {
         return this.data.getUint8(constants.ADDR_SYSTEM_FLAGS) & mask;
     }
 
+    getStartTime () {
+        return this.data.getUint32(constants.ADDR_START_TIME) & mask; // TODO: dataView doesn't support getUint64?
```
RobLoach, line 54: `DataView` doesn't have a `getUint64()` function right now.

devtools/web/src/models/MemoryView.ts
```diff
@@ -44,6 +44,7 @@ export class MemoryView implements MemoryViewComputedProperties {
   readonly pointerPos: Point;
   readonly mouseBtnByte: number;
   readonly systemFlags: number;
+  readonly startTime: number;
```
RobLoach, line 47:
``` suggestion
  readonly time: number;
```

devtools/web/src/models/MemoryView.ts
```diff
@@ -58,6 +59,7 @@ export class MemoryView implements MemoryViewComputedProperties {
     this.palette = extractPalette(dataView);
     this.gamepads = extractGamepads(dataView, bufferedData);
     this.systemFlags = dataView.getUint8(constants.ADDR_SYSTEM_FLAGS);
+    this.startTime = dataView.getUint32(constants.ADDR_TIME); // TODO: DataView doesn't support uint64?
```
RobLoach, line 62:
``` suggestion
    this.time = dataView.getUint32(constants.ADDR_TIME); // TODO: DataView doesn't support uint64?
```

runtimes/native/src/runtime.c
```diff
@@ -23,7 +24,8 @@ typedef struct {
     int16_t mouseY;
     uint8_t mouseButtons;
     uint8_t systemFlags;
-    uint8_t _reserved[128];
+    uint64_t startTime;
```
RobLoach, line 27:
``` suggestion
    uint64_t time;
```

runtimes/native/src/runtime.c
```diff
@@ -53,6 +55,7 @@ void w4_runtimeInit (uint8_t* memoryBytes, w4_Disk* diskBytes) {
     memory->drawColors[1] = 0x12;
     m3ApiWriteMem16(&memory->mouseX, 0x7fff);
     m3ApiWriteMem16(&memory->mouseY, 0x7fff);
+    m3ApiWriteMem64(&memory->startTime, (uint64_t)(time(NULL)));
```
RobLoach, line 58:
``` suggestion
    m3ApiWriteMem64(&memory->time, (uint64_t)(time(NULL)));
```

runtimes/web/src/runtime.js
```diff
@@ -52,6 +52,10 @@ export class Runtime {
         return this.data.getUint8(constants.ADDR_SYSTEM_FLAGS) & mask;
     }
 
+    getStartTime () {
```
RobLoach, line 55:
``` suggestion
    getTime () {
```

JerwuQu:
Might be worth documenting a special case where `TIME` can be set to 0 for MCU-based devices (i.e. arduino) which won't have a system clock, so that games can handle that.

I think something like `INIT_TIMESTAMP` might be more clear than `TIME`, since it both signifies that it will be set once on startup, and that its a unix timestamp rather than something else, but this point is a bit subjective.

For languages that don't support reading u64 yet (DataView should be the only case, right?) it might be OK to just read the lower 32 bits for now. That leaves them safe up until the year 2106, and hopefully they decide to support it before then :)

PR #368, opened 2022-01-29
Support for non-threaded audio, 16-bit graphics and gcc 4.2

phcoder:
* GCC 4.2 is used for osx-ppc and doesn't understand 0b syntax
* PSP hangs with threaded audio
* 16-bit graphics turned out not to be necessarry for PSP support but might be necessary in the future and good to have as a configurable alternative


aduros:
Awesome, thanks!

Issue #195, opened 2021-12-13
Add special characters to system font


aduros:
The [font bitmap](https://github.com/aduros/wasm4/blob/main/runtimes/web/assets/font.png) has empty space reserved for ASCII characters 0x80 through 0xA0.

We should use that space for something, like icons for WASM-4 controller buttons, or generic sprites like stars/hearts/etc.

Games can then use hex string literals to easily display these:

```c
text("Press \x80 to jump!", 0, 0);
```

If anyone enjoys making pixel art and would like to modify the .png, I can follow up with updating the encoded images in the runtimes!

OliveIsAWord:
Here's a first go at possible glyphs.

![image](https://user-images.githubusercontent.com/89336634/147454175-9bc9271b-0107-49af-87e1-ae9d9485b3df.png)

I've posted a prettier, high resolution version in the [Discord](https://discord.gg/7teRdHdbYk). As it stands, these additions would need to be revised, both in the choice of symbols and how they're drawn.

joshgoebel:
https://polyducks.itch.io/kitchen-sink-textmode-font/devlog/222046/kitchen-sink-font

Just game across this...

DenialAdams:
> Here's a first go at possible glyphs.
> 
> ![image](https://user-images.githubusercontent.com/89336634/147454175-9bc9271b-0107-49af-87e1-ae9d9485b3df.png)
> 
> I've posted a prettier, high resolution version in the [Discord](https://discord.com/channels/918503626333356072/918503626333356075/924947572618055720). As it stands, these additions would need to be revised, both in the choice of symbols and how they're drawn.

+1 on playing card suits - would be useful in a project I'm working on right now :)

Issue #366, opened 2022-01-28
Add Controls Icons to Font and Match Current Inputs


desttinghim:
## Problem
Currently each game either uses text or custom graphics to spell out what the controls are, generally by mentioning the keyboard keys to press. This works as long as you are on a PC and have the same keyboard layout. but quickly becomes wrong if you play through retroarch, mobile, or just don't have a standard layout. The current alternative is to use WASM4's naming standard (1, 2, Arrow Keys, etc.), but this is confusing to people who haven't memorized WASM4's buttons and their mappings.

## Proposal
We should add icons to the font that indicate each control (as in #195), and create a system that can swap these icons out with more appropriate icons at runtime. For web/desktop, this means that people can remap the controls to their preference and get correct instructions on what to press. On touch screens we can either add names (1, 2) to the on screen buttons and put this in the icons. In the retroarch builds we can use the retropad as our basis. For implementations of WASM4 on constrained devices this system can be ignored entirely and the icons hard coded into the game.

JerwuQu:
I think this is a great idea, and would kinda be tied into solving #195.

JerwuQu:
Now that we have icons for keyboard, there are a few approaches to switching out the layouts. These are some I came up with:

### Native

#### For keyboard:
Set during runtime depending on keyboard layout. AFAIK has to be pretty much hardcoded to X and Z for QWERTY, J and Q for Dvorak, etc.

#### For libretro and MCUs:
Compile-time flags to set to different common layouts, such as Nintendo-style (A, B), Sony-style (X, O), etc. with an additional flag to swap 1 and 2.

### Web

#### For keyboard:
Same as native.

#### For gamepad overlay:
Either:
1. Label as **1** and **2** and have the font say **1** and **2**.
2. Label as **X** and **Z** and have the same font as when on QWERTY keyboard.

#### For gamepad input:
👻

JerwuQu:
For netplay, an easy fix is to set the layout for each player when the session starts.

Example:
- Player 1: QWERTY: Z and X
- Player 2: Also QWERTY: Z and X

This would actually also help to display the correct layout in the future. Perhaps Player 2 will want to use the input layout for Gamepad 1 (showing arrow-keys and Z and X) despite being represented as Gamepad 2 on netplay.

JerwuQu:
I just realised with the comment above that we would require a section of the font for each gamepad (which in hindsight would be good to have anyways).
`0x80` -> `0x87` for gamepad 1
[...]

PR #369, opened 2022-01-29
Add gamepad icons to font

JerwuQu:
According to #195, this adds the following icons to the font
image
The 1 and 2 icons are by @OliveIsAWord! I took the arrow ones from wasm4nia because I thought it'd be better to have them all look like "buttons". 
They are located at `0x80 + GAMEPAD_BIT`, i.e. `0x80` is `BUTTON_1`, `0x87` is `BUTTON_DOWN`.
The other icons she made could be added too, but I have no clue where (in memory, I mean).


aduros:
I love this, and the layout from 0x80 is smart!

Can we replace 1 and 2 with X and Z? I think it would make more sense to players to use the actual keyboard keys.

JerwuQu:
Yeah, I agree it does make more sense, but only on devices that actually use the keyboard keys X and Z.
I was thinking a bit about #366, but was unsure how to deal with it at this time.

I'm guessing that for the web runtime, it would be fine to call them X and Z, except for when on touch (unless the virtual keypad is also labeled like that). If we want them to be called X and Z on every platform, I think we should also rename the internal names to `BUTTON_X` and `BUTTON_Z`.

A bit conflicted, but I suppose defaulting to X and Z instead of 1 and 2 for now is a better solution and postpones the complexity of #366 a bit.

JerwuQu:
My branch got entirely messed up when rebasing lol
Ignore it for now, will fix.

JerwuQu:
Looks like this now
image

aduros:
Thanks!

Issue #357, opened 2022-01-26
Disk localStorage prefix should be namespaced


aduros:
To prevent different carts from overwriting eachothers data. Maybe use the cart title (`w4 bundle --title`)? Is it worth adding a separate `--storage-key` param to w4 bundle for this?

We also need to handle this for carts embedded on wasm4.org, where the cart slug can be used as a prefix.

PR #371, opened 2022-01-29
web: Add disk localStorage prefix

JerwuQu:
Solves issue #357.

Runtime itself defaults to `disk`.
Bundles default to `<title>-disk` with a CLI option to manually set the prefix.
Site uses `<slug>-disk`.


aduros:
Looks great, thanks!

This will cause current save games to be lost, but that's probably ok.

Issue #367, opened 2022-01-29
Docs: Saving Data questions


jaller94:
I'd find it helpful if the guide answered the following questions about saving data:

* How where is the data being saved?
  * e.g. localStorage in web
  * What about standalone executables?
* Are different games getting different save disks? How are they being told apart?
  * e.g. If multiple games are published on one website, (how) are they not overwriting each other's saves?
* If there are different saves per game, is it possible to upgrade a game?
  * I've started a [Pokemon-like RPG](https://jaller94.gitlab.io/wasm4-monster-collector) and wonder if I will be able to update the game and still continue old saves. (Obviously, the game would need to handle versioning of save formats itself.)

JerwuQu:
> Are different games getting different save disks? How are they being told apart?

This question depends on #357, so we can solve that, then handle all your points in one go!

PR #373, opened 2022-01-30
Answer questions raised in #367

JerwuQu:
#367

You can formulate differently if you wanna, this is just to get the ball rolling :]


jaller94:
How does one choose a game title (or other key) for the save?
  - It the title parameter of w4 bundle, correct?
  - e.g. `npx w4 bundle build/cart.wasm --title 'Monster Collector' --html public/index.html`
For native games does the save get the same name as the cart file?
  - e.g. `monster-collector.exe` creates `monster-collector.sav`?

JerwuQu:
I just realised that maybe this shouldn't get merged before 2.3.0 is released

PR #379, opened 2022-02-03
Fix watch cli

moumar:
hello, i made some improvements and fixes on the "watch" cli command:
- put Makefile detection last in order to be able to have Makefile in supported languages too
- fix filtered directories that caused endless build loop


aduros:
Thanks!

PR #403, opened 2022-02-17
wasm4nia 1.0.4

JerwuQu:
Updated for new APU and uses the WASM-4 font for buttons.

Didn't know if I should update the `date` or not so I left it unchanged. Perhaps there should be two dates for `submitted` and `updated`...? 🤷


aduros:
Thanks! Yeah, we should probably add a second field for a last updated date.

PR #411, opened 2022-02-27
Feature: handle tracef's `%c` as unicode code point

ZetaNumbers:
Previously converted such character to UTF-16 char code,
so large unicode characters would have been truncated.
Now it's possible to pass unicode characters.

### DIFF
--- a/runtimes/web/src/runtime.ts
+++ b/runtimes/web/src/runtime.ts
@@ -318,7 +318,7 @@ export class Runtime {
                     output += "%";
                     break;
                 case 99: // c
-                    output += String.fromCharCode(this.data.getInt32(argPtr, true));
+                    output += String.fromCodePoint(this.data.getInt32(argPtr, true));
                     argPtr += 4;
                     break;
                 case 100: // d


aduros:
We should probably match the same behavior as C's `printf`, which I think truncates to 8 bits for `%c`.

For me this program:

```c
printf("Hello %c\n", 12345678);
```

Prints `Hello N`.

ZetaNumbers:
> We should probably match the same behavior as C's `printf`, which I think truncates to 8 bits for `%c`.

But why? It's not like we are trying to implement libc. With this PR we would able to pass rust's `char` for example.

ZetaNumbers:
Btw if we truncate, should we truncate to 7 bits for ASCII, or truncate to 8 bits and allow some UTF-16 char codes? Aren't non-ASCII characters for `printf` OS dependent?

ZetaNumbers:
> Could we truncate to 8 bits? libc printf semantics aren't perfect, but at least they're well-defined and we don't need to document our own special handling of certain features.

Until and even then we truncate to 8 bits, we probably could handle non-ascii chars as unicode code points instead of UTF-16 char codes?

> For printing unicode characters, isn't it possible to use %s instead of %c? Or just format the string directly in Rust.

ZetaNumbers:
> For printing unicode characters, isn't it possible to use %s instead of %c? Or just format the string directly in Rust.

Current `%s` implementation only works on ascii null-terminated strings.

https://github.com/aduros/wasm4/blob/main/runtimes/web/src/runtime.ts#L272

To manually `tracef` in Rust you would:

1. Create an empty string;
2. Gradually write to this string other substrings, numbers, etc. Meanwhile the `String` would grow (reallocate) gradually increasing its capacity;
3. Flush the whole string onto a single line via `traceUtf8`;
4. Deallocate the string.

This brings some runtime (~7KiB on all code optimizations) into the binary. It could have been better (now only ~2KIB) if there was an ability flush the line by parts, requiring no allocations.

PR #423, opened 2022-03-27
Fix missing rimraf error

voroskoi:
First of all: Thank You for this amazing program :-)

I am not sure if this is the right way to fix it or not, but did work for me.


voroskoi:
Leads to other errors, closing.

PR #432, opened 2022-04-23
Totally not sumo game submit

Andre-LA:
You can play it on itch.io website to see how it looks like:
https://origamisword.itch.io/totally-not-sumo

Built with nelua, wasi-sdk 14 and w4 2.4.0, the netplay works (tested locally with Pop OS and Android).

The png and wasm files are optimized using `pngquant` and `wasm-opt` tools.

Original src code: https://github.com/Andre-LA/totally-not-sumo


aduros:
Thanks!

PR #438, opened 2022-04-29
Add Porth Language support

FrankWPA:
- Adds template for Porth projects
- Adds Porth support to wasm4 CLI
- Adds initial Porth language documentation


cli/assets/templates/porth/Makefile
```diff
@@ -0,0 +1,55 @@
...
+4ORTH_ARGS := -I $(4ORTH_STD) $(4ORTH_ARGS)
+endif
+
+ifeq ($(b), 1)
+4ORTH_ARGS := $(4ORTH_ARGS) -b
+endif
+
+ifeq ($(r), 1)
+4ORTH_ARGS := $(4ORTH_ARGS) -r
+endif
```
aduros, line 25: What do these options do? I wonder if we should have better names for the environment vars?

aduros (approved):
This is so awesome, thank you!

cli/assets/templates/porth/Makefile
FrankWPA: Oh, I did use those to be the same as 4orth compile [options](https://github.com/FrankWPA/4orth#running-options-and-subcommands). We could change the names or put a comment to use `4orth help` to see the command descriptions.
Or we can add aliases as `run` and `bundle`. I'd like to keep the short ones too as it's quicker to type rapidly for a test.

PR #442, opened 2022-05-13
Update wasm4_wasm32.odin

ChuuniMage:
- Update "w4.tone" to take u32s for frequency and duration, to match tutorial API, and to give Odin users the option to feed raw u32s and get the full audio options
- Make w4.tone just use internal_tone, take internal_tone import out of w4.tone_complex so w4.tone can use it


aduros:
Thanks! And cool game 😁

PR #453, opened 2022-05-21
When cart is too big, the error shows how big it is.

jsyrjala:
Changes error message
```
Cart is above the size limit of ${limit} bytes.
```
to
```
Cart size of ${stat.size} bytes is above the limit of ${limit} bytes.
```


aduros:
Thanks!

PR #455, opened 2022-05-23
New game: Touhou Spirits

MrRafael-dev:


aduros:
That's a lot of character art! Did you need to do any fancy compression?

PR #459, opened 2022-06-05
Updating to remove negative checks in Zig example

angeris:
Removing the extraneous negative check from the Zig example in `moving-the-snake.md` as Zig's `@mod` instruction [always returns nonnegative numbers](https://ziglang.org/documentation/master/#mod).


aduros:
Thanks!

PR #463, opened 2022-06-23
Fix typo in Odin's URL 

tupini07:
URLs to the Odin project were prepended with an extra `https://`. This PR fixes them


aduros:
Thanks!

PR #473, opened 2022-07-20
Remove unnecessary import "image" in Go snake tutorial

michaelowens:
In the previous step it explains that, rather than importing Point from image, you create a Point struct to reduce cart size.


aduros:
Thanks!

Issue #470, opened 2022-07-18
Gamepad not working in Firefox in Linux


metiulekm:
Hi,

I couldn't get my gamepad to work on Firefox. It works correctly in <https://gamepad-tester.com>, and it also works correctly when I try running a WASM-4 game in Chrome.

Details:
- Arch Linux with Plasma,
- WASM-4 2.5.0 (AUR package: `wasm4-bin 2.5.0-1`), but the same thing happens on <https://wasm4.org/play>,
- Firefox 102.0.1 (Arch package: `firefox 102.0.1-1`). Sadly I couldn't get WASM-4 to work at all in the Firefox troubleshooting mode, but I did disable extensions and nothing changed,
- gamepad is 8BitDo SN30 Pro+ detected by Gamepad Tester as `Xbox Wireless Controller 045e-02e0-8BitDo SN30 Pro+`,
- the only things I see in the console when running locally are
  ```
  Lit is in dev mode. Not recommended for production! See https://lit.dev/msg/dev-mode for more information. [reactive-element.js:38:20](http://localhost:4444/node_modules/@lit/reactive-element/development/reactive-element.js)
  Multiple versions of Lit loaded. Loading multiple versions is not recommended. See https://lit.dev/msg/multiple-versions for more information. [reactive-element.js:38:20](http://localhost:4444/devtools/web/node_modules/@lit/reactive-element/development/reactive-element.js)
  Warning: Cart is larger than 65536 bytes. Ensure the release build of your cart is small enough to be bundled. [runtime.ts:267:20](http://localhost:4444/src/runtime.ts)
  ```
  (the last one due to testing using `snake-rust`).

Due to my total ineptitude when it comes to webdev I haven't really investigated this too much. One thing that I did manage to find is that `navigator.getGamepads()` works differently across those two browsers: in Chrome it returns `[null, null, null, null]` (as shown by DevTools) before touching the gamepad and `[Gamepad, null, null, null]` afterwards, but in Firefox there are no `null`s: before touching the gamepad it returns `[]` and after touching `[Gamepad]`. This does feel relevant but I don't know.

aduros:
Hi, thanks for the detailed report!

[...]

PR #475, opened 2022-07-21
Add a warning when a non-standard gamepad is detected.

metiulekm:
As discussed in #470.

This will warn once per gamepad (so, if somebody connects an nonstandard gamepad and then tries another one, they will get one warning for each). This needs state in form of the `gamepadUnavailableWarned` variable. I didn't really know if it belongs in `InputState` or not, I ended up putting it in but can move it out if you prefer.

Also, this is my first code in TS in my entire life, so caveat emptor :D I did test it on my gamepad and it does seem to work.

### DIFF
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -382,7 +382,7 @@ export class App extends LitElement {
                     // Let's warn once, and not use this gamepad afterwards.
                     if (!this.inputState.gamepadUnavailableWarned.has(i)) {
                         this.inputState.gamepadUnavailableWarned.add(i);
-                        console.warn("Your browser does not know the mapping for the %s gamepad (index %d) and we can't use it.", gamepad.id, i);
+                        this.notifications.show("Unsupported gamepad: " + gamepad.id);
                     }
                     continue;
                 }
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -14,6 +14,7 @@ import { Notifications } from "./notifications";
 
 class InputState {
     gamepad = [0, 0, 0, 0];
+    gamepadUnavailableWarned = new Set<number>;
     mouseX = 0;
     mouseY = 0;
     mouseButtons = 0;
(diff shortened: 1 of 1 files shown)


runtimes/web/src/ui/app.ts
```diff
@@ -373,9 +374,17 @@ export class App extends LitElement {
...
-                    continue; // Disconnected or non-standard gamepad
+            for (const [i, gamepad] of navigator.getGamepads().entries()) {
+                if (gamepad == null) {
+                    continue; // Disconnected gamepad
+                } else if (gamepad.mapping != "standard") {
+                    // The gamepad is available, but nonstandard, so we don't actually know how to read it.
+                    // Let's warn once, and not use this gamepad afterwards.
+                    if (!this.inputState.gamepadUnavailableWarned.has(i)) {
+                        this.inputState.gamepadUnavailableWarned.add(i);
+                        console.warn("Your browser does not know the mapping for the %s gamepad (index %d) and we can't use it.", gamepad.id, i);
```
aduros, line 385: It would be even nicer to use `this.notifications.show()`, which shows a text notification to the player without them needing to check the JS console:

```suggestion
                        this.notifications.show("Unsupported gamepad: " + gamepad.id);
```

runtimes/web/src/ui/app.ts
```diff
@@ -14,6 +14,7 @@ import { Notifications } from "./notifications";
 
 class InputState {
     gamepad = [0, 0, 0, 0];
+    gamepadUnavailableWarned = new Set<number>;
```
aduros, line 17: What do you think about keying this off the gamepad's ID instead of the index?

I guess that would allow someone to test plugging and unplugging multiple different types of gamepads one at a time and seeing the error message for each invalid one, even if the index is the same.

aduros (commented):
Thanks! I have a couple nitpicks but it looks really solid as is.

runtimes/web/src/ui/app.ts
aduros: (Reduced the amount of text in this error message only because the font for the notifications is pretty huge) 😄

metiulekm: Done.

runtimes/web/src/ui/app.ts
metiulekm: Done.

Though isn't browser supposed to give different gamepads different indices? At least that's how I understood MDN. I guess that it's better this way anyway since the code is a bit simpler.

aduros: From what I understand the index is simply the order that the gamepad was connected, but different gamepads can have the same index if you unplug and replug in a different gamepad.

aduros:
Awesome, thanks!

PR #482, opened 2022-08-03
Fix 3 syntax errors of zig example code

LittleJianCH:


aduros:
Thanks again, it's clear this example code didn't get enough attention 🙂

PR #483, opened 2022-08-04
Added links to user-input guide

thedavesims:
Highlighted that WASM-4 supports two types of user input (gamepad, mouse) with links to the user input guide for both. This will better highlight to new users that mouse input is available.


aduros:
Thanks!

PR #486, opened 2022-08-06
Fix porth `blit` docs

LunaAmora:
Just noticed there was a missing argument on the Porth `blit` documentation.


aduros:
Good catch, thanks!

PR #487, opened 2022-08-07
Update watch.js

lzralbu:
This change solves the following error:

Error: spawn npm ENOENT

Context: Windows 10 with Node.js 16 installed with nvm using Mingw64/MSYS2 terminal.


aduros:
Thanks for this!

PR #501, opened 2022-08-17
Fix `w4 watch` with npm on windows

TheFloatingPixel:
The `w4 watch` command on windows with npm should call the `npm.cmd` file, not `npm`.

Sorry if I did something wrong while creating this pull request, I'm new to GitHub.

### DIFF
--- a/cli/lib/watch.js
+++ b/cli/lib/watch.js
@@ -14,7 +14,7 @@ function start (opts) {
         buildOutput = "target/wasm32-unknown-unknown/debug/cart.wasm";
 
     } else if (fs.existsSync("package.json")) {
-        buildCommand = "npm";
+        buildCommand = process.platform === "win32"? "npm.cmd": "npm";
         buildParams = ["--silent", "run", "build:debug"];
         buildOutput = "build/cart.wasm";
 


aduros:
Hi, thanks for the PR!

@lzralbu Do you have any insights on this? My understanding was that this should be fixed by #487 ? Would applying this change break anything in mingw/msys?

lzralbu:
I usually build with `make` since I develop with C++. That said, I just ran the `w4 watch` command on the AssemblyScript hello world project, which calls `npm`, and it worked just fine, both from command prompt and msys terminal. It would be helpful if @TheFloatingPixel posted the error found.

aduros:
@TheFloatingPixel Can you please confirm you're using the latest version (2.5.2)? This was released very recently, is it possible you could have been using a previous version without the fix from #487 ?

aduros:
@lzralbu If it turns out calling `npm.cmd` is required, would that change still work when running in mingw/msys? I'm not able to test this Windows stuff at the moment.

lzralbu:
> @lzralbu If it turns out calling `npm.cmd` is required, would that change still work when running in mingw/msys? I'm not able to test this Windows stuff at the moment.

Yes, it would. Just tested it.

TheFloatingPixel:
> @TheFloatingPixel Can you please confirm you're using the latest version (2.5.2)? This was released very recently, is it possible you could have been using a previous version without the fix from #487 ?

Just checked, I was on version 2.5.1. The updated version works for me. Sorry for the problem.

PR #505, opened 2022-08-19
Add Gas to devtools view

JerwuQu:


JerwuQu:
I see that there are some conflicts now. @aduros is this PR interesting? If so I could fix them, otherwise I'll close it.

aduros:
Hey, sorry this fell through the cracks. I like the approach you took and think it would make a great opt-in experiment. If you fix the conflicts I'll merge it 🙂 Please also add:

1. Let's always show the gas section on the developer tools. If `--metering` was not used, it should tell the user about the `--metering` option to enable this feature.
2. Add some more documentation, either on the site, in the devtools, or a more descriptive `--metering` help description.

Thanks!

JerwuQu:
I resolved the conflicts for now, but it seems like the `wasm-metering` package itself might be broken for WASM files using newer features. For example, it works on wasm4nia but throws an exception on [Mr.Rafael's Sound Test](https://wasm4.org/play/sound-test). I suspect this is due to bulk-memory extensions.

A closer look and [the repo](https://github.com/ewasm/wasm-metering) is clearly orphaned, so it shouldn't come as a surprise. Either I'd have to fork it and bring it up to date, or we have to put this PR on ice until someone else does so.

PR #511, opened 2022-08-20
Set `touch-action: manipulation` to prevent delays on mobile.

MrRafael-dev:
When playing on mobile, the first touch of any button will freeze the game for a brief moment. This is very annoying for most games.

Using `touch-action: manipulation` seems to fix the delay issue.


aduros:
Seems to work perfectly, thanks a lot!

PR #524, opened 2022-08-23
Panning docs

desttinghim:
Panning is documented in Playing Audio, but not in the function reference. This PR fixes that.


aduros:
Thanks!

PR #525, opened 2022-08-23
Oval fix

### LINKED ISSUE #495: Ovals with even sizes get rounded to odd sizes
Basically, if you were to do `oval(0, 0, 4, 4)`, you'd get a circle 5 pixels wide instead of 4. This scales to any size.

image

desttinghim:
Fixes #495. The problem is that the width and height are using integer bit shifts to divide by two. This lost the least significant bit. I've fixed this by using modulo to get even/odd and offset the line appropriately.  There are still some bugs with oval drawing but I do believe those were already there. I can try and solve those too, but this is what I managed for now.

Here's a small test I made to demonstrate the ovals:

Before:
oval-test-pre

After:
oval-test

```zig

export fn update() void {
    w4.DRAW_COLORS.* = 2;
    const gamepad = w4.GAMEPAD1.*;
    if (gamepad & w4.BUTTON_1 != 0) {
        w4.DRAW_COLORS.* = 4;
    }

    var x: usize = 8;
    var sizex: usize = 3;
    while (x + sizex < 160) : (sizex += 1) {
        var y: usize = 8;
        var sizey: usize = 3;
        while (y + sizey < 160) : (sizey += 1) {
            w4.oval(@intCast(i32, x), @intCast(i32, y), sizex, sizey);
            y += sizey + 1;
        }
        x += sizex + 1;
    }
}
```
[wasm4-oval.zip](https://github.com/aduros/wasm4/files/9400317/wasm4-oval.zip)

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -389,14 +389,20 @@ void w4_framebufferOval (int x, int y, int width, int height) {
     uint8_t strokeColor = (dc1 - 1) & 0x3;
     uint8_t fillColor = (dc0 - 1) & 0x3;
 
+    // Get the width and height
     int a = width >> 1;
     int b = height >> 1;
 
     if (a <= 0) return;
     if (b <= 0) return;
 
+    // Calculate the midpoint
     int x0 = x + a, y0 = y + b;
     int aa2 = a * a * 2, bb2 = b * b * 2;
+
+    // Store even/odd offsets to allow non even ovals
+    int wo = (width + 1) % 2, ho = (height + 1) % 2;
+    int we = width % 2, he = height % 2;
 
     {
         int x = a, y = 0;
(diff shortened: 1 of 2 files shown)


aduros:
Ooh, thank you! I'll check this out.

desttinghim:
I've added a fix for the native runtime as well

MaxGraey:
Nice!
I'm interesting, could this improve to more optimal way (right side)?
ovals

desttinghim:
> Nice! I'm interesting, could this improve to more optimal way (right side)? ovals

Yes, it should be possible, I just need to figure out why it's happening

MaxGraey:
> Yes, it should be possible, I just need to figure out why it's happening

It looks like rounding issue of outline in vertical case and fills and outline on horizontal case

desttinghim:
oval-rewrite-test

I've done a rewrite of the oval algorithm to get  it working consistently. I referenced the TIC-80 ellipse drawing function. Most of the actual math is still going over my head unfortunately :/ I've added comments to aid in understanding what the code is doing and warn that the algorithm has a lot of details to get right.

aduros:
Thanks for working on this! It will be great to have oval drawing fixed.

desttinghim:
@aduros I've made a small change to get the ovals to be less boxy - it was literally just changing 1 to 2 when calculating dy. I'm happy to get this merged whenever you are
oval-rewrite-2-constant

ibillingsley:
> I've made a small change to get the ovals to be less boxy - it was literally just changing 1 to 2 when calculating dy.

Personally I liked the "boxy" ovals better, some sizes look more like diamonds now. In particular 5x5 and 9x9 are used in [my game](https://wasm4.org/play/one-slime-army) which now looks substantially worse.

desttinghim:
> > I've made a small change to get the ovals to be less boxy - it was literally just changing 1 to 2 when calculating dy.
> 
> Personally I liked the "boxy" ovals better, some sizes look more like diamonds now. In particular 5x5 and 9x9 are used in [my game](https://wasm4.org/play/one-slime-army) which now looks substantially worse.

This is good feedback. I'm not opposed to using the more boxy style - I should probably have asked the community which they would have preferred.

PR #530, opened 2022-08-23
Make R reset the cart in native runtime

### LINKED ISSUE #503: Native Runtime: Add `R` key to restart cart

desttinghim:
Closes #503


aduros:
I think this has a memory leak, `w4_runtimeInit` and `w4_wasmLoadModule` aren't currently designed to be called multiple times.

desttinghim:
Oops, I didn't think to check for that. I'll investigate and fix it

desttinghim:
I ran valgrind on this branch and on main, and I don't see any difference in the output. I think the existing memory is being re-used, and nothing is being allocated on reset. Let me know if I'm missing something obvious

aduros:
It's possible that the memory is reused, but that might be undefined behavior. Also, having an explicit teardown method would make it easier to safely port to other wasm engines.

desttinghim:
`w4_runtimeInit` looks like it can be called again no problem - it clears all the memory to zero  and resets everything to default. `w4_wasmLoadModule` sets up the actual runtime, which doesn't need to be repeated, and probably shouldn't be either. Should the teardown function just be a memory reset and then calling `start`?

PR #535, opened 2022-08-24
Update @giscus/react to v1.1.2 (fixes iframe size)

ibillingsley:
Fixes giscus iframe size.

The latest version is 2.2.0, but it has a breaking change so in the meantime this version fixes the iframe sizing issue with minimal impact.


aduros:
Thanks a lot!

Issue #553, opened 2022-09-03
Suggestion: Import/export save data to a file


MrRafael-dev:
Currently, save data is limited to where you play the game:

- Playing on a different site won't share the same save, and it's not possible to move from one location to another, at least not without hacking through developer tools via console.
- If you clear your cookies and history regularly, chances are that your saves are going to be erased, and any progress you've made will have to be done again.
- There's no way to share your progress with others, which can be a fun aspect for completionists, or games with custom maps/characters.
- If you switch between PC and mobile, there's no way of keeping your scores or in-game progress between both of them.
- Games ran through `w4 run` command will share the same save data, which will inevitably cause corruption between different saves.

My suggestion is to have an option to import/export save data to a file, allowing players to share and use them however they like.
Not only this solves a large amount of limitations, it can also provide a lot more if used creatively:

- Save editors for both players and developers.
- If done correctly, save data can be used as an expansion pack, which can provide extra content for games.
- Moving your progress between sequels or other games (ex: carrying your old inventory, or unlocking a secret character after beating another game as a form of easter egg).

aduros:
We have state saving/loading mapped to the `2` and `4` keys. Maybe `Shift-2` can open a dialog to save the state to a file, and `Shift-4` to browse to open?

The state contains both RAM and disk, so it wouldn't be exactly what you're requesting, but might be the simplest and most intuitive for users to understand. Saving the disk alone could be interesting for sharing data between different tools though.

MrRafael-dev:
Having the same option for save states would also be a good addition, but I don't think that should replace the ability of doing the same with disk.

Apart from file sizes, I have a few reasons for this:
- Loading a state on a game with different versions would either revert the game to another version or crash.
- Save editors would be considerably harder to make: instead of only changing a couple of values, now you'll have to deal with the entire RAM.
- Sharing saves between sequels or tools would be impossible.

That said, I think we should still be able to save/load the disk, even if an option for save states already existed.

Maybe disk import/export could have their own options on the menu?
```
Continue
Save state
Load state
Save disk
Load disk
Clear disk
Copy netplay URL
Reset cart
```
We could also have an option to clear disk data, in case we load the wrong save or just want a fresh start.

In regards to saving: I was thinking of just having a single, uncompressed binary file with 1KB of data. Not sure if the file extension could be just `.bin` or should have it's own extension, though.

ZetaNumbers:
[...]

PR #555, opened 2022-09-04
Disk import/export/clear

MrRafael-dev:
Implementation of disk importing/exporting, as initially suggested on issue #553.

Option indexes for menu overlay were moved in order to be a little easier to reorder them if necessary.

There is now a new "*Disk Options*" submenu, which includes four options:
- **Back:** return to previous menu.
- **Export disk:** download the save data to a binary file, called `<prefix>.disk`.
- **Import disk:** open a file picker, select a `.disk` file, then load a game disk.
- **Clear disk:** clear all contents of game disk.

### DIFF
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -581,6 +581,8 @@ export class App extends LitElement {
                         console.error("Error importing disk", error);
                     }
                 }
+
+                app.closeMenu();
             });
 
             reader.readAsArrayBuffer(files[0]);
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -568,10 +568,10 @@ export class App extends LitElement {
                 let disk = new Uint8Array(constants.STORAGE_SIZE);
 
                 disk.set(result);
-                app.runtime.diskBuffer = disk;
+                app.runtime.diskBuffer = disk.buffer;
                 this.runtime.diskSize = result.length;
                 
-                const str = z85.encode(disk);
+                const str = z85.encode(result);
                 try {
                     localStorage.setItem(this.runtime.diskName, str);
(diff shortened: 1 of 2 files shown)


JerwuQu:
Maybe it would be more appropriate to have it be named `<slug>.disk` (or simply `cart.disk`) instead, similar to on the native runtime.

MrRafael-dev:
@JerwuQu Thanks!

I changed the file format to `.disk` and made so it only imports the data being used (disks can be smaller than 1KB). And `localStorage` is also being set/removed after importing/clearing data.

aduros:
Thanks for the PR!

If possible, let's put the disk/state export stuff inside a submenu to avoid the main menu from getting too busy.

Are the `.disk` files cross-compatible with the native runtime?

MrRafael-dev:
@aduros I added a submenu for it. They're all on it's own option, called "*Disk Options*". 👍 

In regards to cross-compatibility: exported `.disk` files have the same format and varying size as the native runtime, so they should be the same.

MrRafael-dev:
Some final fixes:
- When importing a disk, the game will now stay paused until you choose a file.
- You can now reopen the disk file picker after closing it.
- Importing a file will now close the menu.

@aduros I think everything is done now. If you can, please take a look. 😄 

I've been focusing more on disk import/export for this PR. There were some significant changes in regards to how menus work (now we have submenus), so if everything goes well, then other options can be discussed/added later.

aduros:
Thanks! I have some minor tweaks to suggest but nothing blocking.

For example I think we need a default name for disks besides "null" (when running a cart locally with `w4 run`). Also, some of the localStorage stuff is duplicated, I might clean this up a bit later.

aduros:
btw the new possibility of multiple menu screens is exciting!

PR #561, opened 2022-09-09
Oval: restore blockiness

desttinghim:
With the oval fix PR #525 I settled on a less block appearance for the ovals. After it got merged, @ibillingsley commented that they were relying on the blockiness in [their game One Slime Army](https://wasm4.org/play/one-slime-army), and without it the game looks considerably worse. This PR restores the blockiness to the oval drawing.

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -415,7 +415,7 @@ void w4_framebufferOval (int x, int y, int width, int height) {
 
     // Error increments. Also known as the decision parameters
     int dx = 4 * (1 - a) * b * b;
-    int dy = 4 * (b1 + 2) * a * a;
+    int dy = 4 * (b1 + 1) * a * a;
 
     // Error of 1 step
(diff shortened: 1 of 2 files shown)


aduros:
Looks good to me, thanks!

ibillingsley:
[disk-0 MADNESS](https://wasm4.org/play/disk-0-madness/) and [First Flight](https://wasm4.org/play/first-flight/) also looked different for what it's worth. Rather than a subjective decision, I think whichever is more mathematically accurate is best. Could matter if you're trying to do pixel perfect collision detection. I haven't done the math, but all the circle/oval drawing tools I've looked at produce results closer to the original blocky ovals.

ibillingsley:
Actually I think there's still something wrong because now the 9x9 circle is not completely symmetrical .

desttinghim:
Hmm, I should probably make a more comprehensive test cart then. Maybe store the ideal shapes as bitmaps and test that the ovals actually match.

PR #562, opened 2022-09-11
Game Of Life: Zig Edition

davidroman0O:
Here a `zig` version 

I really appreciate `wasm4` so far, good work!

I've cheated by using the frame buffer as the `future` state which turn out pretty well 😄 

The repository is there https://github.com/davidroman0O/wasm4-game-of-life


aduros:
Nice!

PR #563, opened 2022-09-11
Oval fix asymmetry

ibillingsley:
Follow-up to #561.  
I believe I figured out the discrepancy with drawOval compared to the [TIC-80 version](https://github.com/nesbox/TIC-80/blob/e58c583ea7b000446971a311d5469d0b7d90531b/src/core/draw.c#L497) used as reference.

Their parameters are an inclusive bounding box. When calculating `a` and `b` from `width` and `height` we need to subtract `1` to match their algorithm. This seems to correct the asymmetry we were seeing at certain sizes:  
wasm4-screenshot-1

@desttinghim fyi

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -384,7 +384,7 @@ void w4_framebufferRect (int x, int y, int width, int height) {
 // Javatpoint has a in depth academic explanation that mostly went over my head:
 // https://www.javatpoint.com/computer-graphics-midpoint-ellipse-algorithm
 //
-// Draws the eliipse by "scanning" along the edge in one quadrant, and mirroring
+// Draws the ellipse by "scanning" along the edge in one quadrant, and mirroring
 // the movement for the other four quadrants.
 //
 // There are a lot of details to get correct while implementing this algorithm,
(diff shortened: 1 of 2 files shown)


ibillingsley:
For comparison I also tried porting [this algorithm from stackoverflow](https://stackoverflow.com/questions/2914807/plot-ellipse-from-rectangle/54488533#54488533), but I couldn't get the fill quite right. Also it seems much less optimized for performance, but has a slightly rounder look at some sizes.
wasm4-screenshot-2

JerwuQu:
Nice!

I'm kinda wondering why some of the radii seem to give ovals more resembling octagons in the current solution. Comparing to the second one it's very clear. Is this something that could be fixed?

ibillingsley:
Probably. I tried tweaking some of the parameters but wasn't able to improve it without messing up something else. Would take some work to fully understand the algorithm. I think it's partly because of this condition: 
https://github.com/aduros/wasm4/blob/868a6bb1304b3ced3ebd095d7936027e7f0218f8/runtimes/web/src/framebuffer.ts#L211-L212  
If I remove ` || 2 * err > dy` (which seems to prevent doubled pixels in the outline) it looks like this: 
wasm4-screenshot  
but then the vertical and horizontal ones are inconsistent in some cases.

> Comparing to the second one it's very clear.
It's a bit of a mixed bag. I agree the 16x16 circle looks better in algorithm 2, but 14x14 looks better in algorithm 1.

maxcurzi:
I think the latest iteration looks great
image

In the previous the circles have a round shape but the pixels are not drawn with a consistent symmetry (
image

We're splitting hairs (pixels) and the problem here is that there doesn't seem to be a "more correct" way of drawing pixellated circles. https://twitter.com/retronator/status/1029358469185462273/photo/1

In that tweet the top set has circles that are very close to real circles, while the others tend to become octagons...

desttinghim:
I agree that the top set in that tweet looks the best. In a reply the creator of a pixel art software apparently improved their algorithm to make better ellipses.

aduros:
Thank you everyone for looking into this.

aduros:
This seems like an improvement overall, feel free to iterate on it more if anything looks odd. Thanks!

hazeycode:
My little oval shadows and now rectangle shadows :( I guess I should just paint the shape I want though!

PR #569, opened 2022-09-27
Remove incorrect information from `w4 run` docs

majaha:
The docs say that `w4 run` doesn't reload the cart if cart.wasm gets updated while it's running, but it actually does!


aduros:
Thanks!

PR #572, opened 2022-09-28
Added corn.wasm.

SLiV9:
Made with the Penne programming language, as an experiment. I haven't encountered any issues running the game with either web or native runtime (version 2.5.3).


aduros:
Awesome! BTW you asked if PRs for adding support for esolangs would be welcome, the answer is yes 😄 Is Penne at a state where others can easily use it?

SLiV9:
Cool! Not yet. The compiler error messages need an overhaul and the lack of integer casts was a pain point while making this game. But that might be a cool next goal to work towards.

aduros:
Sounds good, for me as long as people can easily install and run it should be good enough to start. It might be a good way to get more users tinkering with it.

PR #573, opened 2022-09-29
Update guides

diogotrentini:
Hi, everyone! :wave:

# Description

After reading all the "Learn" section while testing wasm4 with Rust, I've noticed that some snippets were missing and several `MultiLanguageCode` definitions were commented — probably because the component was currently supporting only a full set of languages.

I've updated the component to support a partial list of languages (with some fallbacks) and also improved the guides a little bit, fixing some typos and adding missing Rust snippets.

# Details about `MultiLanguageCode`

The `MultiLanguageCode` will try to show, in order:

1. The current active language snippet;
2. The AssemblyScript snippet if the current active language snippet is not available; or
3. The first available snippet if the AssemblyScript snippet is also not available.

An example:


site/docs/guides/text.md
```diff
@@ -80,37 +80,41 @@ It's possible to print one of these characters by escaping it's charcode.
...
 
 ```typescript
 // Press UP to jump!
 w4.text("Press \x86 to jump!", 10, 10);
 ```
 
-<!-- </MultiLanguageCode> -->
+```rust
+// Press UP to jump!
+text(b"Press \x86 to jump!", 10, 10);
```
aduros, line 101: Oh, I didn't know about this syntax. Nice!

aduros:
Yeah, this was a big problem with MultiLanguageCode as we've expanded the number of languages, but you've implemented the perfect solution. Thank you!

PR #575, opened 2022-10-05
Add `runner` to Rust's `./.cargo/config.toml`

Cyborus04:
This change allows `cargo run` to run the game in the native runtime automatically, instead of `cargo build & w4 run-native ./target/wasm32-unknown-unknown/debug/cart.wasm`

### DIFF
--- a/cli/assets/templates/rust/.cargo/config.toml
+++ b/cli/assets/templates/rust/.cargo/config.toml
@@ -2,6 +2,7 @@
 target = "wasm32-unknown-unknown"
 
 [target.wasm32-unknown-unknown]
+runner = "w4 run-native"
 rustflags = [
     # Import memory from WASM-4
     "-C", "link-arg=--import-memory",


Cyborus04:
Wait... I tested this on a crate that wasn't generated with `w4 new --rust` 😅
This would also need to change `lib.rs` to `main.rs` and add `#![no_main]`..

aduros:
I'd be fine with renaming `lib.rs`, although I don't fully understand if there would be any other impacts :)

Out of curiosity, would it be better to run with `w4 run` (web runtime with more features) instead of `w4 run-native`?

Cyborus04:
> I'd be fine with renaming lib.rs, although I don't fully understand if there would be any other impacts :)

I can't think of any

> Out of curiosity, would it be better to run with w4 run (web runtime with more features) instead of w4 run-native?

I chose `w4 run-native` simply because it, and in turn `cargo run`, exit immediately when the window is closed, which afaik can't be done with `w4 run` and *has* to be ctrl+c'd out of.

PR #580, opened 2022-10-12
Android: option to vibrate on button press

BigRobCoder:
I haven't made a OSS contribution in a while, and web stuff is not my expertise, so any feedback is appreciated!

### DIFF
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -19,6 +19,13 @@ class InputState {
     mouseButtons = 0;
 }
 
+const vibrateLevels: [string, number][] = [
+    ["OFF", 0],
+    ["LOW", 1],
+    ["MEDIUM", 2],
+    ["HIGH", 4]
+];
+
 @customElement("wasm4-app")
 export class App extends LitElement {
     static styles = css`
@@ -75,6 +82,11 @@ export class App extends LitElement {
     private netplay?: Netplay;
 
     private readonly diskPrefix: string;
+
+    private vibrateLevelIdx = 0;
+    get vibrateLevel (): [string, number] {
+        return vibrateLevels[this.vibrateLevelIdx];
+    }
 
     readonly onPointerUp = (event: PointerEvent) => {
         if (event.pointerType == "touch") {
@@ -664,6 +676,16 @@ export class App extends LitElement {
         netplay.onjoin = playerIdx => this.notifications.show(`Player ${playerIdx+1} joined`);
         netplay.onleave = playerIdx => this.notifications.show(`Player ${playerIdx+1} left`);
         return netplay;
+    }
+
+    canVibrate (): boolean {
+        // No reliable way to tell, so best guess
+        return 'ontouchstart' in window && navigator.vibrate !== undefined;
+    }
+
+    cycleVibrateLevel (): [string, number] {
+        this.vibrateLevelIdx = (this.vibrateLevelIdx + 1) % vibrateLevels.length;
+        return this.vibrateLevel;
     }
 
     getNetplaySummary () {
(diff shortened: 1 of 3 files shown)


runtimes/web/src/ui/app.ts
```diff
@@ -76,6 +83,11 @@ export class App extends LitElement {
 
     private readonly diskPrefix: string;
 
+    private vibrateLevelIdx = 0;
```
BigRobCoder, line 86: I wasn't sure where to store the vibration level. I tried in virtual-gamepad.ts at first, but didn't see how menu-overlay.ts could access it, so it ended up here.

BigRobCoder: I'm defaulting to "OFF" for now, but if feedback is positive I recommend we default to "MEDIUM", I think the vibration really makes it feel more tactile.

runtimes/web/src/ui/menu-overlay.ts
```diff
@@ -115,6 +118,7 @@ export class MenuOverlay extends LitElement {
 
     @state() private selectedIdx = 0;
     @state() private netplaySummary: { playerIdx: number, ping: number }[] = [];
+    @state() private vibrateLevel = "";
```
BigRobCoder, line 121: I tried just referencing app.vibrateLevel, but wasn't seeing render updates immediately. Is this the correct pattern? Mirror the setting in a local `@state()` variabel?

runtimes/web/src/ui/menu-overlay.ts
```diff
@@ -243,11 +258,16 @@ export class MenuOverlay extends LitElement {
     }
 
     render () {
+        if (!this.vibrateLevel) {
+            this.vibrateLevel = this.app.vibrateLevel[0];
```
BigRobCoder, line 262: I tried setting this in the constructor, but looks like `app` is not yet ready, so moved it to here.

BigRobCoder:
Hmm, after some further testing it seems like the vibration stops working after a while. Maybe chrome blocks website vibrations if they are triggered too frequently?

BigRobCoder:
Was testing this for a while tonight on a Samsung Galaxy S21, no issues this time. Tried testing on an iPhone 13 as well, as intended there was no "vibrate" menu option.

aduros:
Thanks so much for the PR! It looks very clean, and this is a great feature.

Some feedback/questions after testing:
- On my device (Android Moto) all the different vibration levels (LOW, MEDIUM, HIGH) feel the same. Is it worth having multiple levels?
- What do you think about only vibrating on touchdown instead of both touchdown and touchup? I don't have a strong opinion here but wondering if that might be more ergonomic?
- Let's avoid calling `navigator.vibrate()` at all if the setting is off, instead of passing it zero.
- In a future PR we could move the vibration setting to a settings submenu, and also persist settings between sessions 😃

BigRobCoder:
> Thanks so much for the PR! It looks very clean, and this is a great feature.
> 
> Some feedback/questions after testing:
> 
> * On my device (Android Moto) all the different vibration levels (LOW, MEDIUM, HIGH) feel the same. Is it worth having multiple levels?

Very interesting. On my Samsung Galaxy S21, the different vibration do feel different, but it probably depends on the device.  The way I have it now, "LOW" should be 1 millisecond of vibration, "MEDIUM" 2, and "HIGH" 4. Maybe I should have more like 10 different levels of vibration, from 1ms to 1024ms?

> * What do you think about only vibrating on touchdown instead of both touchdown and touchup? I don't have a strong opinion here but wondering if that might be more ergonomic?

Yeah, I started out just vibrating on touchdown, but after playing a bunch of games, it just didn't feel right to me, and vibrate both felt more right.

> * Let's avoid calling `navigator.vibrate()` at all if the setting is off, instead of passing it zero.

Sure.

> * In a future PR we could move the vibration setting to a settings submenu, and also persist settings between sessions 😃

Yeah, that makes sense.

PR #582, opened 2022-10-21
Added seed creator showcase cart

huntr3ss:
This cart shows how one can let the player create a random seed that can be used to generate game items (e.g. a map) so that each game can be different (if a different seed is created).

To the best of my knowledge, there is no random number API in wasm4 but the player could add her own seed in order to create a new seed for each individual game (or use the same seed several times).

I have some issues in debug builds. The stack seem to overflow but this issue does not appear in release builds.


aduros:
Thanks! This is a common question, maybe we can link to it in the documentation somewhere?

huntr3ss:
> Thanks! This is a common question, maybe we can link to it in the documentation somewhere?

I fortunately saw the code and could therefore interpret the error regarding the allowed name of the cart. I also did see that cart.js does not exist as described here: https://wasm4.org/docs/tutorials/snake/publish-your-game/#changing-the-list-of-games

Or did I miss it? Especially since I do not see the cart I created on the web page? Or is it not yet build?

Best regards,
Philipp

aduros:
The cart was added, but the timestamp in the readme was older, so it was appearing much further down the list. I updated it here: https://github.com/aduros/wasm4/commit/5a9ea1e19821e55136388861c7cbefe04c076a17 and it should now show up at the top of the site.

huntr3ss:
> The cart was added, but the timestamp in the readme was older, so it was appearing much further down the list. I updated it here: [5a9ea1e](https://github.com/aduros/wasm4/commit/5a9ea1e19821e55136388861c7cbefe04c076a17) and it should now show up at the top of the site.

Thank you very much

Issue #590, opened 2022-11-07
Rust: text function and button characters - the trait bound `[u8; 16]: AsRef<str>` is not satisfied


benc-uk:
Adding the hex codes to show the buttons/arrows results in an error on Rust

```rust
wasm4::text(b"Press \x80 to jump!", 10, 10);
```

Error:
```
error[E0277]: the trait bound `[u8; 16]: AsRef<str>` is not satisfied
   --> src/game.rs:49:25
    |
49  |             wasm4::text(b"Press \x80 to jump!", 10, 10);
    |             ----------- ^^^^^^^^^^^^^^^^^^^^^^ the trait `AsRef<str>` is not implemented for `[u8; 16]`
    |             |
    |             required by a bound introduced by this call
    |
    = help: the following other types implement trait `AsRef<T>`:
              [T; N]
              [T]
    = note: required for `&[u8; 16]` to implement `AsRef<str>`
```

I'm using `rustc 1.65.0 (897e37553 2022-11-02)` on Ubuntu Linux

DenialAdams:
https://doc.rust-lang.org/stable/std/str/fn.from_utf8_unchecked.html

`from_utf8_unchecked` on the byte st will probably work it work, although it is clunky

SLiV9:
Yeah `str::from_utf8_unchecked` is currently the only way to draw the buttons in Rust. Doing so is undefined behavior, but it does seem to work in practice. The underlying issue is that Rust strings have to be valid UTF-8, but WASM-4 expects to be passed strings encoded with ISO-8859-1 (https://github.com/aduros/wasm4/issues/509).

Alternatively you could manually edit `wasm4.rs` by replacing `str` with `[u8]`:
```rust
pub fn text<T: AsRef<[u8]>>(text: T, x: i32, y: i32) {
    let text_ref = text.as_ref();
    unsafe { extern_text(text_ref.as_ptr(), text_ref.len(), x, y) }
}
```
But this would mean that `text("hello", 10, 10)` no longer compiles; you'd need `text(b"hello", 10, 10)`.

PR #591, opened 2022-11-08
Improve Rust template to allow drawing buttons, resolving issue #590

SLiV9:
This pull request resolves issue #590. It is an alternative/predecessor to my earlier (and more involved) pull request #528.

Unlike the other programming languages supported by WASM-4, Rust enforces that its strings (`String`, `&str` and string literals) are valid UTF-8, which prevents game developers from using `\x80` in a string literal to draw the X button. A solution would be using byte strings:
```rust
text(b"Press \x80 to jump!", 10, 10);
```
In fact pull request #573 added this code sample, but it does not compile with v2.5.3 (see #590). I had somehow convinced myself that allowing byte strings would prevent normal string literals (e.g. just `text("Hello world!", 10, 10)`) from compiling, which would have been a big loss in ergonomics for Rust users. That's why I had originally discarded this option. However after commenting on #590 I realized that `str` implements `AsRef<[u8]>`, which means we can use regular strings as long as they only contain US-ASCII, and use byte string literals for special occassions.

This change only affects new Rust crates and is completely backwards compatible.

### DIFF
--- a/cli/assets/templates/rust/src/wasm4.rs
+++ b/cli/assets/templates/rust/src/wasm4.rs
@@ -135,7 +135,7 @@ extern "C" {
 }
 
 /// Draws text using the built-in system font.
-pub fn text<T: AsRef<str>>(text: T, x: i32, y: i32) {
+pub fn text<T: AsRef<[u8]>>(text: T, x: i32, y: i32) {
     let text_ref = text.as_ref();
     unsafe { extern_text(text_ref.as_ptr(), text_ref.len(), x, y) }
 }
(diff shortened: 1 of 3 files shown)


fishtaco567:
In terms of fantasy console feel I certainly prefer this over the more involved proper UTF-8 parsing. Manipulating strings themselves in a minimal rust binary is rather annoying and often requires `from_utf8_unchecked` even in cases where we're not causing UB, so this resolves that friction as well! And of course, getting rid of UB is great!

aduros:
Thanks @SLiV9, this looks like a good solution.

diogotrentini:
>In fact pull request https://github.com/aduros/wasm4/pull/573 added this code sample, but it does not compile with v2.5.3 (see https://github.com/aduros/wasm4/issues/590).

I actually realized I've accidentally sent the code sample on #573 without proposing a change to the Rust template. Thank you for this PR, @SLiV9! Yes, `str` implements `AsRef<[u8]>`, so we should be good here.

PR #598, opened 2022-12-01
Fix out-of-bounds drawing

JerwuQu:
Resulted in different behavior between web and native, and sometimes memory corruption on native.


aduros:
Looks great, thanks!

PR #599, opened 2022-12-01
Add tracef to native runtime

JerwuQu:
**Updated 2024-04-18!**

Using `vprintf` directly would segfault on `%s` because it would try to read the pointer from host memory rather than WASM memory.

Code is mostly ported from web runtime.

## Test code:
```c
#include "wasm4.h"
void start() {
	tracef("|A|%%|%c|%d|0x%x|%s|%f|%u|%d|", 'B', 123, 0xcafe, "string", 1.23, 456, 789);
}
```

## Results
```diff
Old:
- Web:    |A|%|B|123|0xcafe|string|1.23||456|
- Native: |A|%%|%c|%d|0x%x|%s|%f|%u|%d|
New:
+ Web:    |A|%|B|123|0xcafe|string|1.23|%u|456|
+ Native: |A|%|B|123|0xcafe|string|1.23|%u|456|
```

The web runtime was also modified to show unrecognized specifiers as they are rather than remove them (see `%u`).

### DIFF
--- a/runtimes/native/src/runtime.c
+++ b/runtimes/native/src/runtime.c
@@ -169,11 +169,50 @@ void w4_runtimeTraceUtf16 (const uint16_t* str, int byteLength) {
 }
 
 void w4_runtimeTracef (const uint8_t* str, const void* stack) {
-    puts(str);
-
-    // This seems to crash on Linux release builds
-    // vprintf(str, (void*)&stack);
-    // putchar('\n');
+    const uint8_t* argPtr = stack;
+    uint32_t strPtr;
+    for (; *str != 0; ++str) {
+        if (*str == '%') {
+            const uint8_t sym = *(++str);
+            switch (sym) {
+            case 0:
+                return; // Interrupted
+            case '%':
+                putc('%', stdout);
+                break;
+            case 'c':
+                putc(*(char*)argPtr, stdout);
+                argPtr += 4;
+                break;
+            case 'd':
+                printf("%d", *(int32_t*)argPtr);
+                argPtr += 4;
+                break;
+            case 'x':
+                printf("%x", *(uint32_t*)argPtr);
+                argPtr += 4;
+                break;
+            case 's':
+                strPtr = *(uint32_t*)argPtr;
+                argPtr += 4;
+                if (strPtr > 0 && strPtr < sizeof(*memory)) {
+                    printf("%.*s", (int)(sizeof(*memory) - strPtr), (char*)memory + strPtr);
+                } else {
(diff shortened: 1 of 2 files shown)


runtimes/native/src/runtime.c
```diff
@@ -167,11 +167,50 @@ void w4_runtimeTraceUtf16 (const uint16_t* str, int byteLength) {
...
+                } else {
+                    printf("<invalid memory>");
+                }
+                break;
+            case 'f':
+                printf("%lg", *(double*)argPtr);
+                argPtr += 8;
+                break;
+            default:
+                printf("<unknown symbol '%c'>", sym);
```
aduros, line 207: We should probably keep parity with the web runtime and print nothing here. I can see this being more developer friendly though. What do you think?

aduros (commented):
Thanks!

In general what do you feel like should be our strategy with respect to buffer overruns, when users pass in a pointer that doesn't end in a null character? This goes for `trace`, the `%s` param of `tracef`, and `text`.

I guess we have 3 options.

1. Consider these cases undefined behavior.
2. Define that the cart will always crash immediately when an overrun occurs.
3. Cap all reads to the end of the 64 KB memory (perhaps implemented by padding the 64 KB memory with an extra zero byte)

Now that I think about it, we might also want to standardize overrun behavior for `blit`/`blitSub` as well?

runtimes/native/src/runtime.c
JerwuQu: I don't have a strong opinion either way since the trace functions are used more for debugging than gameplay. I do think it's better to print a warning when possible though, but I could add it to the web runtime too in that case.

JerwuQu:
> In general what do you feel like should be our strategy with respect to buffer overruns

Hmm... I want to say option 2 is the best, since crashes are often better than silent weird behavior. This code currently does option 3 since that's what the web runtime does. Perhaps we can split overruns into a separate issue and solve them all at the same time? Pretty much all pointers sent from WASM will have to be bounds- and null-terminator-checked. I'd avoid defining things as undefined behavior since that will make things runtime dependent.

> Now that I think about it, we might also want to standardize overrun behavior for blit/blitSub as well?

Agreed! If we decide on BSOD for strings, we should do it for sprites as well.

JerwuQu:
I changed `<unknown symbol '%c'>` to instead print the unknown character as if it wasn't handled, which should make it just as clear to the user without looking ugly.

For buffer overruns, I will open a separate issue.

aduros:
Thanks!

PR #601, opened 2022-12-02
Remove text null-terminator discrepancy

JerwuQu:
On web, calls to `textUtf8` and `textUtf16` [will stop at a null-terminator](https://github.com/aduros/wasm4/blob/9149980e0f8122fcfa05445baf2d253d991787dc/runtimes/web/src/framebuffer.ts#L277) even when [given a length](https://github.com/aduros/wasm4/blob/main/runtimes/web/src/runtime.ts#L198). Native does not do this, but instead will happily try to print the null character despite there being no font glyph for it.

This commit changes native to also stop at null-terminators to remove the discrepancy between the runtimes.

We could also instead change web to print out the null characters, but that doesn't serve any purpose and is less likely to be what a game creator wants.

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -500,7 +500,7 @@ void w4_framebufferLine (int x1, int y1, int x2, int y2) {
 }
 
 void w4_framebufferText (const uint8_t* str, int x, int y) {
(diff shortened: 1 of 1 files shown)


aduros:
Sounds good to me, thanks.

PR #603, opened 2023-01-06
Documentation: Make end frequency begin at 1Hz.

datMaffin:
The end frequency can not be 0, because this is a special value reserved whether a end frequency exists.

I know, a bit of a nitpick ;)


datMaffin:
TBH: I am not even sure if changing this is really a good idea. Because the "end frequency" is still zero. But a "end frequency" of zero is special. So maybe this should not be changed.

PR #606, opened 2023-01-15
Add Penne language support

SLiV9:
- Added a hello world template for Penne v0.3.0
- Added `--penne` support to _w4 new_, _w4 watch_ and _w4 png2src_
- Added Penne code samples to the docs

The quick start guide currently mentions WASI SDK, but the Makefile only uses `$WASI_SDK_PATH/bin/clang`. It should be possible to use stock `clang`, but I haven't been able to get it to call `wasm-ld` correctly, whereas WASI SDK worked out of the box.


aduros:
Exciting! Thanks for this!

PR #611, opened 2023-02-21
Fix #417

### LINKED ISSUE #417: `tracef` does not format arguments when running using `w4 run-native`
`tracef` does not format arguments when running using `run-native`. 

Example code:

```c
#include "wasm4.h"

void start() {
  tracef("String: %s", "Ok");
  tracef("Int: %d", 42);
}
```

Example output with `w4 run`:
```
String: Ok
Int: 42
```

Example out with `w4 run-native`:
```
String: %s
Int: %d
```

Edit: 
I wrote this issue before I found [the commit responsible](https://github.com/aduros/wasm4/commit/429ede2589a5bfda18c7b86e07f0d5850a462eb1), I feel it's still worthwhile to have an issue to track it if that's okay?

jacereda:
This is my attempt to fix `tracef` for the wasm3 runtime. Looks convoluted, and it is, but it's just that implementing this portably is a PITA.

### DIFF
--- a/runtimes/native/src/backend/wasm_wasm3.c
+++ b/runtimes/native/src/backend/wasm_wasm3.c
@@ -151,7 +151,7 @@ static m3ApiRawFunction (traceUtf16) {
 static m3ApiRawFunction (tracef) {
     m3ApiGetArgMem(const char*, str);
     m3ApiGetArgMem(const void*, stack);
-    w4_runtimeTracef(str, stack);
+    w4_runtimeTracef(str, stack, _mem);
     m3ApiSuccess();
 }
 
--- a/runtimes/native/src/backend/wasm_wasm3.c
+++ b/runtimes/native/src/backend/wasm_wasm3.c
@@ -151,7 +151,7 @@ static m3ApiRawFunction (traceUtf16) {
 static m3ApiRawFunction (tracef) {
     m3ApiGetArgMem(const char*, str);
     m3ApiGetArgMem(const void*, stack);
-    w4_runtimeTracef(str, stack, _mem);
+    w4_runtimeTracef(str, stack);
     m3ApiSuccess();
 }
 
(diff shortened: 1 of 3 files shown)


JerwuQu:
Also see #599

jacereda:
That one is simpler, but doesn't allow formatting (%08d). Also, it seems to emit a newline at the end, that's bad. `tracef` should behave like `printf`. Otherwise there's no way to emit a partial line, and that's quite useful.

JerwuQu:
It behaves in both of those ways to be identical to the web runtime.

jacereda:
In that case, the web runtime is seriously broken. There must be some way to emit a single character to the console without a newline.

JerwuQu:
I personally don't see this is a big issue since you are quite literally able to format the string and shouldn't need multiple calls to it. What is the use-case here that you feel this strongly about?

More technically, web browsers can't print single characters to the log, so you'd have to buffer it. If doing it that way, aside from breaking all current games using `tracef`, it's arguably also unintuitive to new users when nothing comes out in the console at all if they forget the newline.

LunaAmora:
> I personally don't see this is a big issue since you are quite literally able to format the string and shouldn't need multiple calls to it. What is the use-case here that you feel this strongly about?

The use case I feel strongly about this, is that when experimenting with creating toy Langs, it puts some unecessary difficulty/annoyance on debugging stuff, as you would have to implement an string buffer first to have an clearer debug experience. But this is not a common thing to need, so i know it's not that important :p

jacereda:
It's important to me because that's exactly my use case, a toy language.

How about having a SYSTEM_STANDARD_TRACING to opt-in for this behaviour in the wasm3 runtime?

JerwuQu:
Since #599 got merged, could this PR be closed in favor of the issue with the no-newline case?

aduros:
Yeah, let's close this one and use #612 to discuss what to do about single character printing.

PR #614, opened 2023-02-24
Improve piping support and add `-` alias.

LunaAmora:
- [Fixes `w4 run` piping](https://github.com/aduros/wasm4/commit/1d1f2b7c6b1da2bce2dacdf16da64d12f0e0afc8)
   - Waits for the end of the input instead of starting the server automatically without the cart.
- [Adds piping support to the native backend](https://github.com/aduros/wasm4/commit/19074ad63358cd8a72cd941623e4477c6bd024dd)
   - Avaiable from both the executable and the `w4 -run-native` cli option
- Adds `-` as an alias for `/dev/stdin` for both backends.
   - Both should work on windows, but was not tested

### DIFF
--- a/cli/lib/server.js
+++ b/cli/lib/server.js
@@ -5,7 +5,7 @@ const path = require("path");
 const qrcode = require("qrcode");
 const { Server: WebSocketServer } = require("ws");
 const open = require("open");
-const { exit } = require("process");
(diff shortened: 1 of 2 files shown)


aduros:
Thanks!

PR #621, opened 2023-03-07
Fix explanation of DRAW_COLORS

StijnRuts:
Order of fill and outline was swapped


aduros:
`DRAW_COLORS` is indexed low bits to high bits, so the example of `*DRAW_COLORS = 0x42` is read right-to-left: color 1 is 2 and color 2 is 4, color 3 is 0, and color 4 is 0.

This may seem kinda backwards at first, but it allows you to omit the high bits, eg: just `*DRAW_COLORS = 0x2` to change the fill color of rectangles and everything else transparent.

StijnRuts:
I see, that makes sense.
The use of 'first' and 'second' color is a bit ambiguous tough. I have added the clarification of low/high bits.

PR #628, opened 2023-03-25
Improve crashes and bluescreening

majaha:
This pull request simplifies the bluescreening logic, and fixes some bugs where errors were getting swallowed when the cart failed to instantiate by re-throwing them.


aduros (commented):
Thanks! Good changes for the most part, but let's keep the `start` and `update` exports optional instead of raising an error.

majaha:
Ah yes, good point!

majaha:
I've pushed the fix.

PR #629, opened 2023-03-25
fix compilation for Zig 0.11.0

### LINKED ISSUE #619: Zig v0.11 breaks w4's build file
Relevant post: https://devlog.hexops.com/2023/zig-0-11-breaking-build-changes/

Version 0.11 of Zig (still in pre-release) introduced breaking changes to the build system. Now, running `zig build` on the files generated by w4 puts out this error:

```
error: no field or member function named 'standardReleaseOptions' in 'Build'
```

Though you might want to wait until the stable release before trying to fix this, it's possible they might make more changes. "The breakings will continue", as the lead Zig dev once said 👀

wooster0:
This makes the w4 template for Zig work for me. I think the majority of breaking changes are done and it might stay like this until 0.11 is released. You might want to merge this only when 0.11 is actually released, though, so I marked this PR as a draft because things might still change.

Fixes #619

### DIFF
--- a/cli/assets/templates/zig/build.zig
+++ b/cli/assets/templates/zig/build.zig
@@ -16,7 +16,7 @@ pub fn build(b: *std.Build) !void {
     lib.stack_size = 14752;
 
     // Export WASM-4 symbols
-    lib.rdynamic = true;
+    lib.export_symbol_names = &[_][]const u8{ "start", "update" };
 
     lib.install();
 }
--- a/cli/assets/templates/zig/build.zig
+++ b/cli/assets/templates/zig/build.zig
@@ -18,5 +18,5 @@ pub fn build(b: *std.Build) !void {
     // Export WASM-4 symbols
     lib.export_symbol_names = &[_][]const u8{ "start", "update" };
 
-    lib.install();
+    b.installArtifact(lib);
 }
--- a/cli/assets/templates/zig/build.zig
+++ b/cli/assets/templates/zig/build.zig
@@ -1,18 +1,22 @@
 const std = @import("std");
 
-pub fn build(b: *std.build.Builder) !void {
-    const mode = b.standardReleaseOptions();
-    const lib = b.addSharedLibrary("cart", "src/main.zig", .unversioned);
+pub fn build(b: *std.Build) !void {
+    const optimize = b.standardOptimizeOption(.{});
 
-    lib.setBuildMode(mode);
-    lib.setTarget(.{ .cpu_arch = .wasm32, .os_tag = .freestanding });
+    const lib = b.addSharedLibrary(.{
+        .name = "cart",
+        .root_source_file = .{ .path = "src/main.zig" },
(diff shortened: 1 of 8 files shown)


cli/assets/templates/zig/build.zig
```diff
@@ -1,18 +1,22 @@
...
-    lib.setBuildMode(mode);
-    lib.setTarget(.{ .cpu_arch = .wasm32, .os_tag = .freestanding });
     lib.import_memory = true;
     lib.initial_memory = 65536;
     lib.max_memory = 65536;
     lib.stack_size = 14752;
 
     // Export WASM-4 symbols
-    lib.export_symbol_names = &[_][]const u8{ "start", "update" };
+    lib.export_symbol_names = &[_][]const u8{ "update" };
```
aduros, line 19: Is removing "start" here intentional?

aduros:
Thanks for this!

cli/assets/templates/zig/build.zig
r00ster91: Yes, I got a compile error with it included. `start` is already exported by default.

r00ster91: Ok I've done some research and here's what I've found:
1. We could either do it the current way `lib.export_symbol_names = &[_][]const u8{ "start", "update" };` and require the user to have `export fn start() void {}` in their  code if they have no `start` code.
2. We could do `lib.rdynamic = true;` which will actually include all exported functions in the Wasm binary. I think this is the best approach because it doesn't require the user to have `export fn start() void {}` in their code or even know about the existence of `start` to make their game run. I don't think it'll be a problem to always include all functions that are explicitly marked as `export` in the Zig code.

aduros: Requiring the `export` keyword on the `start` and `update` functions sounds fine to me.

r00ster91: Done

luehmann (commented):
I think you'll also need to remove the no longer present `--prominent-compile-errors` flag from https://github.com/aduros/wasm4/blob/950ad1cdf1829cc0415d6b4ce27dcd2b448a85c6/cli/lib/watch.js#L23

ibillingsley:
Zig 0.11.0 is now released: https://ziglang.org/download/0.11.0/release-notes.html

[wasm4.zig](https://github.com/aduros/wasm4/blob/main/cli/assets/templates/zig/src/wasm4.zig) in the template also needs a small update, `@intToPtr` → `@ptrFromInt`

See https://github.com/ibillingsley/wasm4-gamejam/commit/8d5771c21cbc88ab777c64fd5c62ecd652cf24b9#diff-ab4b0a94f8632ca185834ce5d0090b3342a6772579dcaf8cc42b1de37aaaf514

aduros:
I haven't gotten a chance to test this personally, but the PR looks good to me!

PR #630, opened 2023-03-26
Remove duplicate instructions in web and native runtime's `oval()`

grantshandy:
While fixing [wasmstation](https://github.com/wasmstation/wasmstation)'s oval() I found that there are quite a few little "twice-defined values" in this function.

This PR fixes:
 - `err2` being calculated twice.
 - `b * b` and `a * a` being calculated five separate times.
 - `a` being modified in a way that's inconsistent with how `b1` is modified next to it.
 
 in both the web and native runtimes.
 
 In theory, my changing `err * 2 > dy` to `err2 > dy` would result in different behavior because `err` may be modified after `err2` is defined. In practice, I haven't seen this happen. I ran both the web and native runtimes with this change through [wasm4-test-cart](https://github.com/JerwuQu/wasm4-test-cart/) and there was no difference in all **~1500** oval() assertions.

This change should create identical ovals with fewer operations.

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -413,27 +413,34 @@ void w4_framebufferOval (int x, int y, int width, int height) {
                             // one (overlapping the top line) for even heights
 
     // Error increments. Also known as the decision parameters
-    int dx = 4 * (1 - a) * b * b;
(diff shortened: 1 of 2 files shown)


aduros:
Nice catch, thanks!

Issue #570, opened 2022-09-27
`w4 run` crashes if the cart file is deleted


majaha:
Running `w4 run` and then deleting the cartridge while it's running causes a crash.
For example, I ran `rm output/cart.wasm` during this run:
```
Matt@Matts-PC ~/p/w/w/wave-cycle (master)> w4 run --no-qr output/cart.wasm
Open http://localhost:4444. Press ctrl-C to exit.
✔ Reloaded cart.wasm
events.js:291
      throw er; // Unhandled 'error' event
      ^

Error: ENOENT: no such file or directory, open 'C:\cygwin64\home\Matt\prog\web\wasm4\wave-cycle\output\cart.wasm'
Emitted 'error' event on ReadStream instance at:
    at internal/fs/streams.js:136:12
    at FSReqCallback.oncomplete (fs.js:156:23) {
  errno: -4058,
  code: 'ENOENT',
  syscall: 'open',
  path: 'C:\\cygwin64\\home\\Matt\\prog\\web\\wasm4\\wave-cycle\\output\\cart.wasm'
}
Matt@Matts-PC ~/p/w/w/wave-cycle (master) [1]>
```
Which is a little annoying, as I like to delete all the stale build artefacts as part of my build process.
It would be nice if `w4 run` continued to watch the path and load the new file if one appears there.

PR #632, opened 2023-03-28
Fix several bugs in server.js, including #570

majaha:
Fix the server crashing if the cart file is deleted (issue https://github.com/aduros/wasm4/issues/570)
Fix the server trying to re-read standard in when trying multiple ports.
Fixes several other minor bugs.

Also included, you can take or leave this one: some fun emoji in the cart-reload message 😀

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -125,7 +125,7 @@ withCommonRunOptions(program.command("run <cart>"))
     .addOption(
         new Option("--settle-time <time>", "Changes to the cart must have stopped for at least this long before a reload happens. Increase this if you're getting erroneous double-reloads and cart corruption, decrease it if you want snappier reloads.  In milliseconds.")
         .env("W4_SETTLE_TIME")
-        .default(500)
+        .default(300)
     )
     .action((cart, opts) => {
         const server = require("./lib/server");
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -122,6 +122,11 @@ withCommonRunOptions(program.command("watch"))
 
 withCommonRunOptions(program.command("run <cart>"))
     .description("Open a cartridge in the web runtime")
+    .addOption(
+        new Option("--settle-time <time>", "Changes to the cart must have stopped for at least this long before a reload happens. Increase this if you're getting erroneous double-reloads and cart corruption, decrease it if you want snappier reloads.  In milliseconds.")
+        .env("W4_SETTLE_TIME")
+        .default(500)
+    )
     .action((cart, opts) => {
         const server = require("./lib/server");
         server.start(cart, opts);
--- a/cli/lib/server.js
+++ b/cli/lib/server.js
@@ -1,126 +1,182 @@
 const express = require("express");
-const fs = require("fs");
+const fs = require('node:fs/promises');
 const os = require("os");
 const path = require("path");
 const qrcode = require("qrcode");
 const { Server: WebSocketServer } = require("ws");
 const open = require("open");
-const { exit, stdin } = require("process");
+const process = require("process");
+const { buffer } = require('node:stream/consumers');
 
-var attempts = 0;
-var PORT = 0;
-var FIRST_PORT = 0;
 
-function start (cartFile, opts) {
+async function start (cartFile, opts) {
     const app = express();
-    setupCart(app, cartFile).then(() => {
-        startServer(app, opts, cartFile)
-    });
-}
 
-async function setupCart(app, cartFile) {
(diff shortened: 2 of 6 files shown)


majaha:
Hmm, I'm having issues running the packaged up binary version of w4. When I run `w4 run` I get:
<details><summary>pkg error</summary>

```
pkg/prelude/bootstrap.js:1872
      throw error;
      ^

Error: Cannot find module 'fs/promises'
Require stack:
- C:\snapshot\cli\lib\server.js
- C:\snapshot\cli\cli.js
1) If you want to compile the package/file into executable, please pay attention to compilation warnings and specify a literal in 'require' call. 2) If you don't want to compile the package/file into executable and want to 'require' it from filesystem (likely plugin), specify an absolute path in 'require' call using process.cwd() or process.execPath.        
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:815:15)
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1951:46)
    at Function.Module._load (internal/modules/cjs/loader.js:667:27)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at Module.require (pkg/prelude/bootstrap.js:1851:31)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (C:\snapshot\cli\lib\server.js:2:12)
    at Module._compile (pkg/prelude/bootstrap.js:1926:22)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'C:\\snapshot\\cli\\lib\\server.js', 'C:\\snapshot\\cli\\cli.js' ],     
  pkg: true
}
```
</details> 

It looks like my use of [node:fs/promises](https://nodejs.org/api/fs.html#file-system) broke this, but
I'm not sure how to fix it. Maybe it's a bug in [pkg](https://github.com/vercel/pkg)?

majaha:
Looks like it's all good to go 👍

cli/lib/server.js
```diff
@@ -1,126 +1,185 @@
...
-                    const cart = path.basename(cartFile);
-                    if (hot) {
-                        console.log(`✔ Hot swapped ${cart} (press R for full reload)`);
-                    } else {
-                        console.log(`✔ Reloaded ${cart}`);
-                    }
+    const hot = opts.hot;
+
+    if (shouldWatch) {
+        console.log(opts.settleTime);
```
aduros, line 103: Stray log 🙂

cli/cli.js
```diff
@@ -122,6 +122,11 @@ withCommonRunOptions(program.command("watch"))
 
 withCommonRunOptions(program.command("run <cart>"))
     .description("Open a cartridge in the web runtime")
+    .addOption(
+        new Option("--settle-time <time>", "Changes to the cart must have stopped for at least this long before a reload happens. Increase this if you're getting erroneous double-reloads and cart corruption, decrease it if you want snappier reloads.  In milliseconds.")
+        .env("W4_SETTLE_TIME")
+        .default(500)
```
aduros, line 128: 500ms seems a bit long here. Have you been seeing errors with lower values?

cli/lib/server.js
```diff
@@ -1,126 +1,185 @@
...
-async function getStdinBuffer() {
-	const result = [];
-	var length = 0;
+async function wait (ms) {
+    return new Promise((resolve, reject) => setTimeout(resolve, ms));
+}
+
+// Watches a file-path, and calls handleFreshFile when the file is replaced or modified,
+// once all changes have settled down.
+async function watch_path(path, pollTime, handleFreshFile) {
```
aduros, line 131:
```suggestion
async function watchPath(path, pollTime, handleFreshFile) {
```

Style nitpicks, could you please use `camelCase` instead of `snakey_case` when working in the JS/TS parts of the codebase?

As an aside, I think it'd be good to beef up our linting rules to make styling a little more uniform, although I'm not a stickler for code formatting.

aduros (changes requested):
Good changes here, thanks!

cli/cli.js
majaha: I agree that keeping this short is important, but I was seeing stray reloads and occasional corruption when it was set at 250ms, I think caused by the separate `wasm-opt` step. I thought it was better to have a safe default and let people with fast machines tune it down. I've just tested again, and on my setup 300ms seems pretty stable. I'll change it to that.

PR #640, opened 2023-04-28
Hotfix PING to enable multiplayer

CanyonTurtle:
@aduros woopsie! I just realized PING multiplayer was disabled from a bug, here's a fix.


CanyonTurtle:
Thanks for getting this merged! When will it go live?

PR #647, opened 2023-05-30
Make the invite link work for html bundles

leighmcculloch:
### What
Make the invite link the same URL that the game is being displayed from with the netplay peer ID parameter, instead of a link to wasm4.org/netplay.

### Why
When bundling wasm4 games as HTML, it's an odd user experience that invite links take players off to another website.

### DIFF
--- a/runtimes/web/src/netplay/index.ts
+++ b/runtimes/web/src/netplay/index.ts
@@ -249,7 +249,14 @@ export class Netplay {
     }
 
     getInviteLink (): string {
-        var url = new URL(window.location.href);
+        const loc = window.location;
+        if (loc.protocol == "file:"
+            || loc.hostname == "localhost"
+            || loc.hostname == "127.0.0.1"
+            || loc.hostname == "wasm4.org") {
+            return `https://wasm4.org/netplay/#${this.peerMgr.localPeerId}`;
+        }
+        var url = new URL(loc.href);
         url.searchParams.set('netplay', this.peerMgr.localPeerId);
         return url.href;
     }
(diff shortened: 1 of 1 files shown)


leighmcculloch:
@aduros What do you think?

leighmcculloch:
@aduros If there's an alternative way you'd like addressing this issue approached, I'm happy to have a go of that too.

aduros:
This sounds good to me, but I think we need a special case for the games hosted on https://wasm4.org/play.

leighmcculloch:
What's the best way to signal that to the code generating the bundle?

leighmcculloch:
A command line option?

aduros:
I don't think we need a new param. The easiest way might be to branch on `location.host`/`location.protocol` in `getInviteLink`. If hosted on file://*, wasm4.org, or localhost it should use the original wasm4.org/netplay link, otherwise it can use the same URL with the netplay param.

leighmcculloch:
That's a trivial change. I can make that.

leighmcculloch:
I've made the change so that the original URL is preserved for file, localhost, and wasm4.org. On other domains it uses the existing URL.

aduros:
PR looks great, but is the new link not working on the connecting side? Nothing should have changed on the backend. Can you share a link? If you think it was just a network hiccup, I can just merge this 🙂

leighmcculloch:
It must be my network, because the netplay on wasm4.org with games hosted there is broken for me too. But it was working a couple weeks ago.

leighmcculloch:
@aduros Wdyt, mergeable?

aduros:
Sounds good! Thanks.

PR #648, opened 2023-05-30
Detect mobile with hover: none

leighmcculloch:
### What
Detect mobile with hover: none instead of pointer: coarse.

### Why
pointer: coarse is not a reliable way to identify mobile touch devices. Android Pixel 7 devices report themselves as pointer: fine.

### DIFF
--- a/runtimes/web/src/ui/virtual-gamepad.ts
+++ b/runtimes/web/src/ui/virtual-gamepad.ts
@@ -28,7 +28,7 @@ export class VirtualGamepad extends LitElement {
         :host {
             display: none;
         }
-        @media (pointer: coarse) {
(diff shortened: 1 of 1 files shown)


majaha:
Here's a relevant table in the spec: https://drafts.csswg.org/mediaqueries-4/#mf-interaction

It seems that by the spec, `pointer: coarse` should be sufficient. However, this seems to be broken on some Pixel phones at the moment, here's some relevent bug trackers:
https://github.com/ionic-team/ionic-framework/issues/26240
https://bugs.chromium.org/p/chromium/issues/detail?id=1384181

aduros:
Thanks!

PR #652, opened 2023-07-02
Resolving issue with copying netplay link on safari

BeefOnWeck:
Current fallback for copying netplay url to clipboard does not work in safari due to restriction that it must occur within a user interaction event. Proposed solution is to create a dialog box from which the user can copy the link and then close the dialog. Added benefit is that the current fallback makes use of execCommand, which is deprecated.

### DIFF
--- a/runtimes/web/src/ui/utils.ts
+++ b/runtimes/web/src/ui/utils.ts
@@ -42,25 +42,29 @@ export async function copyToClipboard (text: string): Promise<void> {
         // Attempt to use the async clipboard API, which is not universally supported
         await navigator.clipboard.writeText(text);
     } catch {
-        // Fall back to using execCommand on a dummy textarea element
-        const textArea = document.createElement('textarea');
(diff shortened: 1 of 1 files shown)


leighmcculloch (approved):
🎉

aduros:
Is there anything we can do to make sure the clipboard copy happens in a user interaction event?

BeefOnWeck:
This is what the dialog looks like with iOS Safari:
Maze Racer

Since every browser on the iPhone is webkit under the hood, players currently cannot get the netplay link if they're using an iPhone. The workaround is to start the game on another platform that has access to a non-webkit browser. This PR makes it so that iPhone users can at least get the netplay link.

aduros:
I think the root issue is the pause menu button tap is on pointerdown, whereas iOS only considers pointerup to be a user interaction. We can probably fix this by waiting for the following pointerup after pointerdown to perform the copy.

I'll merge this for now though as a workaround until we can look into that. Thanks!

PR #660, opened 2023-08-20
Fix the annotations in the Zig Snake tutorial

ziebam:
Without the asterisks (`*`) I encountered some memory issues when using Zig's `std.BoundedArray`. Since I never used the language, it took me a while to figure out the problem, so I'm updating it for the future users.


aduros:
Thanks!

Issue #664, opened 2023-09-30
Start function is not called when DEV_NETPLAY flag is set.


JohnLunsford:
When the DEV_NETPLAY flag is set to true, copyNetplayLink() is called in runtimes/web/app.ts. This sets this.netplay = true and copies the neplay url to the clipboard. However, this call to copyNetplayLink() is currently happening before the call to runtime.start(), which is only called if this.netplay == false. Since copyNetplayLink() sets this.netplay = true, this prevents runtime.start() from being called. It makes sense to avoid calls to runtim.start() by non-host players who join the game. However, it needs to be called by the host in order to perform any initial setup the game may require. Moving the check for the DEV_NETPLAY flag and the call to copyNetplayLink() after the call to runtime.start() fixes this issue.

PR #665, opened 2023-09-30
proposed fix for issue #664

JohnLunsford:
When the DEV_NETPLAY flag is set to true, copyNetplayLink() is called in runtimes/web/app.ts. This sets this.netplay = true and copies the neplay url to the clipboard. However, this call to copyNetplayLink() is currently happening before the call to runtime.start(), which is only called if this.netplay == false. Since copyNetplayLink() sets this.netplay = true, this prevents runtime.start() from being called. It makes sense to avoid calls to runtim.start() by non-host players who join the game. However, it needs to be called by the host in order to perform any initial setup the game may require. Moving the check for the DEV_NETPLAY flag and the call to copyNetplayLink() after the call to runtime.start() fixes this issue.

### DIFF
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -131,10 +131,6 @@ export class App extends LitElement {
             this.netplay.join(hostPeerId);
         } else {
             await runtime.load(await loadCartWasm());
-
-            if (DEV_NETPLAY) {
-                this.copyNetplayLink();
-            }
         }
 
         let devtoolsManager = {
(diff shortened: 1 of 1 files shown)


majaha:
Makes sense to me.

Although I wonder if `copyNetplayLink()` should be renamed. Perhaps `maybeCreateNetplayAndCopyLink()`.
Or maybe it should be changed so that it just checks if a netplay instance exists and copys it's link, without trying to create one if there isn't.

aduros:
Nice catch, thanks!

JohnLunsford:
np! Btw, I’m really loving wasm4. Thanks so much for creating it!On Oct 3, 2023, at 1:52 AM, Bruno Garcia ***@***.***> wrote:﻿
Nice catch, thanks!

Issue #649, opened 2023-06-05
WASM globals are not managed by the netplay synchronization engine


DenialAdams:
WASM globals are not managed by the netplay synchronization engine. As a result of this, two players on a cart using WASM globals can go out of sync.

We are at the mercy of compilers here, hoping that they choose to persistently store data in memory rather than in WASM global variables.

It is unknown (to me) how widespread the use of WASM globals is in compiler backends. There is at least one (Roland) that would run into this issue were it not for a [WASM-4 specific check](https://github.com/DenialAdams/roland/commit/b2e77b4b742074baf44d22e7f5cffc8cc21b773c).

DenialAdams:
One possible solution for this issue could be a cart pre-processing step that marks all global variables as exported, so that the WASM-4 runtime can observe them.

JerwuQu:
We are now syncing globals with the help of a preprocessing step like you suggested 😄 Can this be closed?

DenialAdams:
Awesome! Thank you for your work!

PR #666, opened 2023-09-30
Always call runtime.reset() in App.resetCart()

majaha:
I think this is technically the right thing to do, just in case the wasm start function or WASI _start/_initialize functions read from memory and then set some global variables (I don't know why they would, but they could.)
Probably wouldn't be necessary if we [fixed the global variable problem](https://github.com/aduros/wasm4/issues/649), and saved them in the `State`? Although thinking about it, there's other global state like wasm tables that we should worry about.

(Aside: I think we should try to tidy up the abstraction between runtime.reset(), runtime.load() and App.resetCart() a little, they seem a bit inter-dependant.)

Sorry for the train-of-thought pull request, I just find it helpful to have things written down 😊

### DIFF
--- a/runtimes/web/src/ui/app.ts
+++ b/runtimes/web/src/ui/app.ts
@@ -646,9 +646,9 @@ export class App extends LitElement {
             // Take a snapshot
             state = new State();
             state.read(this.runtime);
-        } else {
-            this.runtime.reset(true);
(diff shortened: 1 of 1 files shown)


aduros:
Makes sense! Thanks for digging into this.

PR #673, opened 2023-11-17
print wrappers with comptime formatting and type safety

Ruulul:
add functions `print`, that prints on console, `printText`, that wraps the `text` function, and `printBuffer`, that prints on an arbitrary buffer. allows leveraging custom formatters on custom types, type safety, and use the error handling for ergonomics.

### DIFF
--- a/cli/assets/templates/zig/src/wasm4.zig
+++ b/cli/assets/templates/zig/src/wasm4.zig
@@ -134,3 +134,31 @@ extern fn traceUtf8(strPtr: [*]const u8, strLen: usize) void;
 /// See https://github.com/aduros/wasm4/issues/244 for discussion and type-safe
 /// alternatives.
 pub extern fn tracef(x: [*:0]const u8, ...) void;
+
+/// zig powered formatted print with type safety at compile time.
+/// prints on debug console.
+/// `extra_len` specifies how many characteres you will need at most, _on addition_ to the format string length.
+/// You can pass negative numbers, as long as the value is comptime known.
+pub fn print(comptime extra_len: comptime_int, comptime fmt: []const u8, args: anytype) !void {
+    var str = [_]u8{0} ** (@as(comptime_int, fmt.len) + extra_len);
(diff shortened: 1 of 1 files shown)


xdBronch:
the stdlib already has [functions to print to a buffer](https://ziglang.org/documentation/master/std/#A;std:fmt.bufPrint) and with an arguably better API. im also a little confused about the other functions youve added, if i know the `extra_len` that i want, why not just create an array myself and use the buffer print function?

Ruulul:
again didnt know about this one handy function

about the why... I just realized it was being too common to write out a block, create a scoped buffer, print to that buffer, then send the buffer to the proper function (`w4.trace` or `w4.text`)

and on a second note, I dont like to need to manually count characters to know the needed buffer size, and it is not viable to do the std.debug trick and make a `[1024]u8` buffer on wasm4, so I settled to an offset based on the fmt string len

mostly I wanted a way to be able to do this kind of pattern in an ergonomic way:
```zig
w4.textPrint(1, "Gen:\n{}", 135, 10, .{generation}) catch {
  w4.text("Gen:\nXe?", 135, 20);
};
```

I am curious tho of what is the expected way to handle printing in general... without a wrapper I would need to do something like...
```zig
{
  const fmt = "Gen:\n{}";
  var buffer: [@as(comptime_int, fmt.len) + 1]u8 = undefined;
  w4.text(try std.fmt.bufPrint(&buffer, fmt, .{generation}) catch "Gen:\nXe?", 135, 20);
}
```

I understand tho if you see this as not general enough for the wasm4.zig file. I just thought it is a pretty common need to want to use a formatter, and it is unpleasant to need to manually count characters to know the proper offsets

Ruulul:
I just realized I can write a simple comptime logic to make all calls use the same buffer, but I am not sure if this is worthywhile

JerwuQu:
I think this is better handled by users since you both need to specify how big you want the buffer to be and it adds a dependency on `std`.

PR #669, opened 2023-11-07
zig: Replace `addSharedLibrary` with `addExecutable`

peterhellberg:
> [!IMPORTANT]
> Related to the breaking changes in https://github.com/ziglang/zig/pull/17815

This requires `0.12.0-dev.1482+e74ced21b` or later (currently the latest `master` release on https://ziglang.org/download/)

PR #677, opened 2023-11-26
2 simple fixes

Inve1951:
b19c86b15bc09c1bbdfbda3a003b92b61297d005: Fixes `w4` looking for `cart.wasm` in the wrong location since #669.
5063e90d7fc592fae5a1168cbfd19ca7473a697d: Makes the native binary compatible with linux installations older than ~2 years.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,8 @@ jobs:
                       wasm4: bin/wasm4
                       libretro: lib/wasm4_libretro.dylib
 
-                    - os: ubuntu-latest
+                    # glibc 2.31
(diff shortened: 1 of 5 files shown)


Inve1951:
I saw #675 only after this.
If you wanna merge that one I'll drop my equivalent commit and rebase.

aduros:
Thanks for the Linux tweak. Are there any caveats we should watch out for?

Inve1951:
> Thanks for the Linux tweak. Are there any caveats we should watch out for?

You're welcome. No caveats. This just lowers the minimum library version required at runtime.

PR #680, opened 2023-12-13
Passing correct FPS into devtools

JohnLunsford:
The correct delta time for the frame was not being passed into the devtoolsManager.updateCompleted() call. (This apparently was a known issue since there is a comment that says "FIXME(2022-08-20): Pass the correct FPS for display."  This commit adds a variable to track a timestamp for the last completed frame update and passes the delta between this timestamp and the current frame timestamp into the updateCompleted() call. 

The resulting displayed FPS is accurate, but can be choppy due to the intermittent frame delays that are caused by waiting for timeFrameStart to catch up to timeNextUpdate. So, to smooth this out, this commit also includes a change to devToolsManager to display a rolling 10 frame average FPS.

### DIFF
--- a/devtools/web/src/devtools-manager.ts
+++ b/devtools/web/src/devtools-manager.ts
@@ -58,6 +58,20 @@ export class DevtoolsManager {
    */
   private _bufferedData = new BufferedRuntimeData();
 
+  private _fpsBuffer : number[] = [0,0,0,0,0,0,0,0,0,0];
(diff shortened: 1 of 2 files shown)


JohnLunsford:
FIXES #681

JohnLunsford:
Noticed an issue with the average calculation. Will post a new pull request that includes the fix

PR #682, opened 2023-12-13
Pass correct fps to devtools

### LINKED ISSUE #681: Correct FPS not being passed to devtools
The correct delta time for the frame is not being passed into the devtoolsManager.updateCompleted() call. Instead, a hard-coded value of 1000/60 is being passed. As a result, the devtools do not display the actual FPS for the game. (This apparently was a known issue since there is a comment that says "FIXME(2022-08-20): Pass the correct FPS for display."

JohnLunsford:
FIXES #681 

The correct delta time for the frame was not being passed into the devtoolsManager.updateCompleted() call. (This apparently was a known issue since there is a comment that says "FIXME(2022-08-20): Pass the correct FPS for display." This commit adds a variable to track a timestamp for the last completed frame update and passes the delta between this timestamp and the current frame timestamp into the updateCompleted() call.

The resulting displayed FPS is accurate, but can be choppy due to the intermittent frame delays that are caused by waiting for timeFrameStart to catch up to timeNextUpdate. So, to smooth this out, this commit also includes a change to devToolsManager to display a rolling 10 frame average FPS.

### DIFF
--- a/devtools/web/src/devtools-manager.ts
+++ b/devtools/web/src/devtools-manager.ts
@@ -58,6 +58,20 @@ export class DevtoolsManager {
    */
   private _bufferedData = new BufferedRuntimeData();
 
+  private _fpsBuffer : number[] = [0,0,0,0,0,0,0,0,0,0];
+  private _fpsBufferIdx = 0;
+  //calculate an average FPS for the last 10 frames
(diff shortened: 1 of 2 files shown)


aduros:
Thanks!

PR #686, opened 2024-01-08
Add support for Roc in png2src

bhansconnect:
A friend and I have been working on support w4 in Roc. That project lives here: https://github.com/lukewilliamboswell/roc-wasm4/

This adds png2src support for Roc. Note, Roc does not store all flags with a sprite. Instead, roc just stores bits per pixel and leaves the rest of the flags for when calling `blit` or `blitSub`.

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -4,7 +4,7 @@ const { program, Option } = require("commander");
 const pkg = require('./package.json');
 const { supportedIconExtensions } = require('./lib/utils/icon');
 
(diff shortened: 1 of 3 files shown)


aduros:
Nice, thanks! Would you be interested in writing additional documentation to the website?

bhansconnect:
> Nice, thanks! Would you be interested in writing additional documentation to the website?

I am not sure that would be the best idea with the current state of Roc. Roc is still a very young language and is not yet to a numbered release. There is a very good chance that docs added today could be partially stale in a few months.

I really hope that Roc reaches it's first numbered release by the end of the year, but currently I know there are a number of breaking changes on the horizon. That said, definitely interested in adding docs overall.

Issue #685, opened 2024-01-07
Zig template doesn't build


squeaktoy:
after doing `w4 new --zig test` and trying `zig build -Doptimize=ReleaseSmall` I get this:
```
/home/user/my_software/test4/build.zig:8:14: error: no field named 'cpu_arch' in struct 'Build.ResolvedTarget'
            .cpu_arch = .wasm32,
             ^~~~~~~~
/usr/lib64/zig/9999/lib/std/Build.zig:2242:28: note: struct declared here
pub const ResolvedTarget = struct {
                           ^~~~~~
referenced by:
    runBuild__anon_8077: /usr/lib64/zig/9999/lib/std/Build.zig:1851:37
    steps__anon_7900: /usr/lib64/zig/9999/lib/build_runner.zig:1033:29
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
```

And even with Zig 0.11.0 it fails:
```
/home/user/my_software/test4/build.zig:14:9: error: no field named 'entry' in struct 'Build.Step.Compile'
    exe.entry = .disabled;
        ^~~~~
/usr/lib64/zig/0.11.0/lib/std/Build/Step/Compile.zig:1:1: note: struct declared here
const builtin = @import("builtin");
^~~~~
referenced by:
    runBuild__anon_7138: /usr/lib64/zig/0.11.0/lib/std/Build.zig:1639:37
    remaining reference traces hidden; use '-freference-trace' to see all reference traces
```

Removing the `exe.entry = .disabled;` line makes matters even worse:
```
zig build-exe cart ReleaseSmall wasm32-freestanding: error: the following command failed with 1 compilation errors:
/usr/lib64/zig/0.11.0/bin/zig build-exe --stack 14752 /home/user/my_software/test4/src/main.zig -OReleaseSmall --cache-dir /home/user/my_software/test4/zig-cache --global-cache-dir /home/user/.cache/zig --name cart --import-memory --initial-memory=65536 --max-memory=65536 --export=start --export=update -target wasm32-freestanding -mcpu generic --listen=-
Build Summary: 0/3 steps succeeded; 1 failed (disable with --summary none)
install transitive failure
└─ install cart transitive failure
   └─ zig build-exe cart ReleaseSmall wasm32-freestanding 1 errors
/usr/lib64/zig/0.11.0/lib/std/start.zig:559:45: error: root struct of file 'main' has no member named 'main'
[...]

PR #692, opened 2024-01-17
zig: Fix template build

mrBen:
I'm not sure when or why these changes occurred in Zig, but with this fix I can compile the template on the latest version of Zig.


mrBen:
I don't know how to link it, but it fixes #685.

aduros:
Thanks, it's hard to keep on top of all these breaking Zig changes 😬

PR #693, opened 2024-02-05
Update Dependencies, fix type errors, and general improvements.

majaha:
I updated the dependencies, most notably Vite to version 5. I also reviewed how Vite was being configured and fixed development builds being used in production.

I also fixed several typescript issues, the code now passes type-checking.

There are also various other little tidy-ups and improvements.


aduros:
This is amazing, thanks a lot!

PR #694, opened 2024-02-13
Fix imports in the WAT template

orsinium:
1. `diskr` and `diskw` return an i32 (number of bytes written).
2. `tracef` is not provided by the host environment, it's just a helper in the C SDK.

You didn't have an error running the example because binaryen strips out all the unused imports. It would fail if you use any of these function or if you don't have binaryen installed.


aduros:
Thanks!

PR #705, opened 2024-04-16
✨ Make APU tick aware ✨ 

JerwuQu:
## What

It has always been an issue that if you try to continously play `tone` each frame, you will get gaps due to frame timing issues. This has made doing live-playback or creating very fancy sounds difficult.
The previous workaround has been to extend each `tone` so that the next one will override it, but this has several drawbacks when working with pitch-bends.

What is done here is make the APU aware of WASM-4 ticks. This enables us to keep playing the active `tone` past its expiration as long as the next tick hasn't rolled in yet, thus avoiding gaps in playback.

## Demo

Here is a video demonstration of the fix: [**demo**](https://github.com/aduros/wasm4/assets/3710677/34f58e2f-1d12-4a8a-ade9-6af052062b34)

This also shows that it only affects continously calling `tone` with sustain and doesn't affect intentional skipping or envelopes (which is correct.)
Has been tested on both web and native, on Windows and Linux. 

## Example

Here's the cart I used to test this and the Zig code for it: [**cart.wasm**](https://github.com/aduros/wasm4/files/15002323/cart.zip)

```zig

export fn start() void {}

var ticks: usize = 0;
export fn update() void {
    w4.DRAW_COLORS.* = 0x14;
    w4.text("\x80:continous", 5, 5);
    w4.text("\x81:skip 1", 5, 15);
    w4.text("\x84:skip 2", 5, 25);
    w4.text("\x86:continous release", 5, 35);
    w4.text("\x85:skip 1 release", 5, 45);

    w4.DRAW_COLORS.* = 0x13;
    if (w4.GAMEPAD1.* & w4.BUTTON_1 > 0) {
        w4.text("continous", 5, 80);
        w4.tone(500, 1, 50, 0);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_2 > 0) {
        if ((ticks % 2) == 0) {
            w4.text("skip 1", 5, 80);
            w4.tone(500, 1, 50, 0);
        }
    } else if (w4.GAMEPAD1.* & w4.BUTTON_LEFT > 0) {
        if ((ticks % 3) == 0) {
            w4.text("skip 2", 5, 80);
            w4.tone(500, 1, 50, 0);
        }
    } else if (w4.GAMEPAD1.* & w4.BUTTON_UP > 0) {
        w4.text("continous release", 5, 80);
        w4.tone(500, 1 << 8, 50, 0);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_RIGHT > 0) {
        if ((ticks % 2) == 0) {
            w4.text("skip 1 release", 5, 80);
            w4.tone(500, 1 << 8, 50, 0);
        }
    }
    ticks += 1;
}
```

I also tested some code doing continous frequency sweeps and the difference is night and day, even when doing workarounds with the old version.

### DIFF
--- a/runtimes/native/src/apu.c
+++ b/runtimes/native/src/apu.c
@@ -29,6 +29,9 @@ typedef struct {
     /** Time the tone should end. */
     unsigned long long releaseTime;
 
+    /** The tick the tone should end. */
+    unsigned long long endTick;
+
     /** Sustain volume level. */
     int16_t sustainVolume;
 
@@ -59,8 +62,9 @@ typedef struct {
 
 static Channel channels[4] = { 0 };
 
-/** The current time, in samples. */
+/** The current time in samples and ticks respectively. */
 static unsigned long long time = 0;
+static unsigned long long ticks = 0;
 
 static int w4_min (int a, int b) {
     return a < b ? a : b;
@@ -71,6 +75,7 @@ static int lerp (int value1, int value2, float t) {
 }
 
 static int ramp (int value1, int value2, unsigned long long time1, unsigned long long time2) {
+    if (time >= time2) return value2;
     float t = (float)(time - time1) / (time2 - time1);
     return lerp(value1, value2, t);
 }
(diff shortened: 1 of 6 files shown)


aduros:
Looks great, thanks!

PR #711, opened 2024-04-19
apu: fix continous tone for triangle

JerwuQu:
Related to #705 

I found that the triangle wave behaved incorrectly, and this was due to its specific `releaseTime` being added.
I had tested the triangle wave locally before submitting the last PR, but did so at a different sample rate so maybe that affected something, I don't know.

Here's a piece of code to test this bug with:

```zig
const w4 = @import("wasm4.zig");

export fn start() void {}

var ticks: usize = 0;
export fn update() void {
    w4.DRAW_COLORS.* = 0x14;
    w4.text("\x80:short attack", 5, 5);
    w4.text("\x81:short decay", 5, 15);
    w4.text("\x84:short sustain", 5, 25);
    w4.text("\x86:long release", 5, 35);
    w4.text("\x87:short sus tri", 5, 45);
    w4.text("\x85:short decay tri", 5, 55);

    w4.DRAW_COLORS.* = 0x13;
    if (w4.GAMEPAD1.* & w4.BUTTON_1 > 0) {
        w4.text("A", 5, 80);
        w4.tone(60, 1 << 24, 50 | (50 << 8), 0x40);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_2 > 0) {
        w4.text("D", 5, 80);
        w4.tone(60, 1 << 16, 50 | (50 << 8), 0x40);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_LEFT > 0) {
        w4.text("S", 5, 80);
        w4.tone(60, 1, 50 | (50 << 8), 0x40);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_UP > 0) {
        w4.text("R", 5, 80);
        w4.tone(60, 255 << 8, 50 | (50 << 8), 0x40);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_DOWN > 0) {
        w4.text("D tri", 5, 80);
        w4.tone(60, 1 << 16, 50 | (50 << 8), 0x40 | 0x2);
    } else if (w4.GAMEPAD1.* & w4.BUTTON_RIGHT > 0) {
        w4.text("S tri", 5, 80);
        w4.tone(60, 1, 50 | (50 << 8), 0x40 | 0x2);
    }
    ticks += 1;
}
```

As a bonus, also contains a fix majaha mentioned if someone would call `tone` in `start` but not export an `update` from their WASM.

Issue #678, opened 2023-12-01
Inconsistent Audio Output


Inve1951:
On my machine WASM-4 produces inconsistent audio output.

Simple example cart:

```zig
const w4 = @import("wasm4.zig");
var tick: usize = 0;
export fn start() void {}
export fn update() void {
    tick += 1;
    switch (tick) {
        10 => {
            w4.tone(262, 60, 50, 8);
        },
        70 => {
            w4.tone(262, 60, 50, 8); // square
            w4.tone(262, 60, 50, 2); // triangle
        },
        130 => {
            w4.tone(262, 60, 50, 2);
        },
        else => {},
    }
}
```

About 1 in 3 times when loading (or reloading) this cart it plays completely different audio.
I can reproduce this issue in firefox and linux native. I have not tried other runtimes.

Audio Sample (linux native):

https://github.com/aduros/wasm4/assets/18194808/a445f1d7-a9d1-4774-b070-adb43d131274

I'm not sure what's going on here. Especially the middle note seems affected. Sometimes the distortion is subtle and sometimes it's very noticable, with a bit of randomness.

Pre-built Cart + HTML + Linux Binary:
[binary.zip](https://github.com/aduros/wasm4/files/13520426/binary.zip)

WASM4 Version:
yesterday's master + my unrelated patches from #677 ([history](https://github.com/aduros/wasm4/commits/5063e90d7fc592fae5a1168cbfd19ca7473a697d))

JerwuQu:
This sounds like phase shenaningans (google phase cancellation), and the difference in audio when running the same cart would be because the phase of the channels aren't always consistent. The reason the differences happens is likely because rendering updates, meaning calls to `update` and hence calls to `tone`, don't happen on the same audio sample number every time you start WASM-4 (an inherent flaw in running the audio engine based on an inaccurate timer).

I'm not an emulator developer, but here are some fixes I can think of and what I think their downsides are.

The simplest fix for this particular case of variance would be to reset the phase to 0 every time a `tone` is triggered. This has a pretty nasty downside of giving audible pops if re-triggering a note though, since the phase gets shifted mid-sound. This is most clearly heard with the triangle channel, and I think there even was an issue/PR for it way back.

Another (albeit more difficult) fix would be to rewrite the audio engine's oscillators to go based on sample number for phase rather than having a separate `phase` variable for each channel. This would keep the same relative phase between channels, but the notes can still start at different points in the phase. I personally have no clue if this is a good idea.

[...]

PR #714, opened 2024-04-28
Fix triangle release by switching period by tick

JerwuQu:
Fixes the bug with the 1ms triangle release not happening after #711.

As you can see, it's not as clean as one could hope.

This is kind of a compromise between having tone be sample-based and having it tick-based. I believe this is the best kind of compromise when perfect 60 Hz ticks aren't possible though. Basically, we respect samples when it comes to envelopes A->D->S but respect ticks for starting the Release.
In practice this means that if ticks *were* to drift (which they almost always do), they only do so during the Sustain period. If we were to make all parts of ADSR tick based, the drift could happen at any parts. 

Still, this is a bit unfortunate for the case of frequency slides where we can't know how many samples long a tone will be until we have already detected the drift, meaning we either have to end the slide short (because of tick-sample drift), or we have to adjust it as it happens (leading to pitch skips).
I chose here to go with having the slide end short which would be the least audible of the two. This required introducing a new `releaseTime` for what is estimated when the tone starts playing to be able to have a target time for `getCurrentFrequency`.

I am thinking even more about rewriting the WASM-4 `update` to be based on the audio routine to both fix the tick-sample drift and issue #678 for good, but for now this is what we can do.

Since this is a bit more code, I would appreciate someone else checking the logic so I don't introduce any new bugs!

### DIFF
--- a/runtimes/native/src/apu.c
+++ b/runtimes/native/src/apu.c
@@ -7,7 +7,7 @@
 #define MAX_VOLUME 0x1333 // ~15% of INT16_MAX
 // The triangle channel sounds a bit quieter than the others, so give it higher amplitude
 #define MAX_VOLUME_TRIANGLE 0x2000 // ~25% of INT16_MAX
-// Also the triangle channel prevent popping on hard stops by adding a 1 ms release
+// Also for the triangle channel, prevent popping on hard stops by adding a 1 ms release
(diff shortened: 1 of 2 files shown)


majaha:
Heya. I've been working on my own patches over here: https://github.com/aduros/wasm4/pull/716 which I think come at some of the problems you were attacking with a simpler approach. Could you try those changes out and see if they solve the issue you're trying to fix here?

Issue #542, opened 2022-08-28
Netplay not synced when using start function


1j01:
I was initializing my game inside the `start` callback, but this lead to a memory access error when trying to use netplay, or if I defined empty arrays at declaration like `let players: Player[] = [];` then the game would render but with entities missing on the client player's side. The host would see all the entities as well as inputs from the client player, but the client's game was broken.
I fixed it by moving initialization out of `start` and doing it at top level.
Is `start` just not supported yet for netplay?
This is a very cool feature, by the way! I love how simple you're trying to make it.

aduros:
Hi, thanks a lot for reporting this!

I believe `start` shouldn't be called for joining clients, since calling it after joining would usually clobber the game state they just received from the host. There might actually be a bug causing `start` to be called when it shouldn't be.

aduros:
Cool game btw! I love Nidhogg.

CanyonTurtle:
I ran into a desync issue and fixed it because of this thread. Thanks for documenting this! 

My fix: https://github.com/CanyonTurtle/PING/commit/a045f986e3c0f27a34498ac40e999beca41c2af8#diff-d924ca21d81d7d5a59eb9e10d3e2689c4c58a7e28e6ecce6a064701666f5a730

My game: 
https://canyonturtle.github.io/PING/

P.S. let me know what you think @aduros of the game! I'm not done with it yet but it's up there.

JerwuQu:
I built the game from the linked commit and the issue is clearly present. 
[Here's the built cart.wasm + cart.wat](https://github.com/aduros/wasm4/files/15016107/build.zip) for anyone else interested in taking a look.

To elimiate the case of `start` being called multiple times, I added some debug logging in the runtime, and at least that is not the issue here.

I checked a `wasm2wat` conversion of the WASM and there are 26 WASM globals which immediately raises red flags, but since moving init stuff from `start` into `update` fixes the issue, I don't see how that could actually affect anything.

I managed to break at the first `update` in the joining client and it looks like the globals are indeed different:
![2024-04-17_21-38-02_Discord_RpdLcufGfX](https://github.com/aduros/wasm4/assets/3710677/7c3ea90d-8cd6-4149-8151-f493aaa546db)
Not entirely sure this is the issue because the workaround works, but I wouldn't rule it out...

Tricky! Would be nice to test again if we ever fix globals for netplay (is it even possible?)

JerwuQu:
[...]

PR #715, opened 2024-04-28
netplay: patch export globals and sync them

JerwuQu:
Solves #649 and #542.

Tested with the cart from https://github.com/aduros/wasm4/issues/542#issuecomment-2062089238.

I understand if this is not the direction you wanna go and maybe something should be done differently, but it's a start at finally getting this fixed :)

### DIFF
--- a/runtimes/web/src/runtime.ts
+++ b/runtimes/web/src/runtime.ts
@@ -4,6 +4,7 @@ import { APU } from "./apu";
 import { Framebuffer } from "./framebuffer";
 import { WebGLCompositor } from "./compositor";
 import * as devkit from "./devkit";
+import { wasmPatchExportGlobals } from "./wasm-patch";
 
 export class Runtime {
     canvas: HTMLCanvasElement;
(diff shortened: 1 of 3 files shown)


aduros:
I like the idea of preprocessing the wasm to export globals, but using JSON in the serialization seems inefficient.

I wonder if we can assume that the entire globals section is a known fixed size and ordering. Then we'd only need to write the values into the saved state. I think we could also write each value as binary, the exact type also being known. What do you think?

JerwuQu:
Unfortunately I don't think we can get globals as binary data from JS and would have to convert it back and forth from JS types ourselves. At that point I'd guess the browser implementation of JSON isn't actually slower and is a lot less code.

I'd agree with looking into fancier solutions if we serialized this often, but to my knowledge it's only on client connections and user savestates (?)

Feel free to look into it! I'm pretty happy with the WASM patcher code but for the State-code I just made something that works :)

aduros:
> I'd agree with looking into fancier solutions if we serialized this often, but to my knowledge it's only on client connections and user savestates (?)

For netplay we serialize the state every frame, and rollbacks can result in multiple unserialize/serializes in a single frame. Not a huge deal, but it would be good to optimize later 🙂 Let's merge it for now to unblock wasm languages that rely on globals.

PR #716, opened 2024-05-01
Various fixes and improvements to the APU

majaha:
After the [tick-awareness was added](https://github.com/aduros/wasm4/pull/705) recently, I took the opportunity to review the APU code, and found various issues I could fix.

The biggest changes are that now tone() calls will all take affect simultaneously at the end of a tick, and that the audio will pause more smoothly when the game is paused. Other changes are mostly bug fixes and clean-ups.

### DIFF
--- a/runtimes/native/src/runtime.c
+++ b/runtimes/native/src/runtime.c
@@ -40,6 +40,13 @@ static Memory* memory;
 static w4_Disk* disk;
 static bool firstFrame;
 
+static MaybeToneCall bufferedToneCalls[4] = {
+    {false, 0, 0, 0, 0},
+    {false, 0, 0, 0, 0},
+    {false, 0, 0, 0, 0},
+    {false, 0, 0, 0, 0},
+};
+
 void w4_runtimeInit (uint8_t* memoryBytes, w4_Disk* diskBytes) {
     memory = (Memory*)memoryBytes;
     disk = diskBytes;
@@ -128,7 +135,12 @@ void w4_runtimeTextUtf16 (const uint16_t* str, int byteLength, int x, int y) {
 
 void w4_runtimeTone (int frequency, int duration, int volume, int flags) {
     // printf("tone: %d, %d, %d, %d\n", frequency, duration, volume, flags);
-    w4_apuTone(frequency, duration, volume, flags);
+    int channelIdx = flags & 0x3;
+    bufferedToneCalls[channelIdx].active = true;
+    bufferedToneCalls[channelIdx].frequency = frequency;
+    bufferedToneCalls[channelIdx].duration = duration;
+    bufferedToneCalls[channelIdx].volume = volume;
+    bufferedToneCalls[channelIdx].flags = flags;
 }
 
 int w4_runtimeDiskr (uint8_t* dest, int size) {
@@ -223,7 +235,7 @@ void w4_runtimeUpdate () {
         w4_framebufferClear();
     }
     w4_wasmCallUpdate();
-    w4_apuTick();
+    w4_apuTick(bufferedToneCalls);
     uint32_t palette[4] = {
         w4_read32LE(&memory->palette[0]),
         w4_read32LE(&memory->palette[1]),
--- a/runtimes/web/src/apu-worklet.ts
+++ b/runtimes/web/src/apu-worklet.ts
@@ -5,6 +5,8 @@ const SAMPLE_RATE = 44100;
 const MAX_VOLUME = 0.15;
 // The triangle channel sounds a bit quieter than the others, so give it higher amplitude
 const MAX_VOLUME_TRIANGLE = 0.25;
+// The small fade-out time channels use to reduce audio popping.
+const FADE_OUT_TIME = Math.floor(SAMPLE_RATE / 200);
 class Channel {
 
     // Tone Parameters
@@ -88,6 +90,9 @@ class Channel {
 
     /** The last generated random number, either -1 or 1. */
     noiseLastRandom = 0;
+
+    /** The time remaining until the fade out period ends. */
+    fadeOutTime = 0;
 }
 
 function lerp (value1: number, value2: number, t: number) {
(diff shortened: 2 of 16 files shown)


runtimes/native/src/runtime.c
```diff
@@ -128,7 +135,12 @@ void w4_runtimeTextUtf16 (const uint16_t* str, int byteLength, int x, int y) {
 
 void w4_runtimeTone (int frequency, int duration, int volume, int flags) {
     // printf("tone: %d, %d, %d, %d\n", frequency, duration, volume, flags);
-    w4_apuTone(frequency, duration, volume, flags);
+    int channelIdx = flags & 0x3;
+    bufferedToneCalls[channelIdx].active = true;
+    bufferedToneCalls[channelIdx].frequency = frequency;
+    bufferedToneCalls[channelIdx].duration = duration;
+    bufferedToneCalls[channelIdx].volume = volume;
+    bufferedToneCalls[channelIdx].flags = flags;
```
JerwuQu, line 143: Not sure if keeping APU state outside of the APU is desired here.

The APU already has internal global state, so maybe this logic could be placed in `apu.c` instead to keep it isolated?

JerwuQu (commented):
Aside from the comment, I will give this a try to see if the approach fixes the problems I've seen previously!

runtimes/web/src/apu-worklet.ts
```diff
@@ -117,7 +114,7 @@ class APUProcessor extends AudioWorkletProcessor {
 
     getCurrentVolume (channel: Channel) {
         const time = this.time;
-        if (time >= channel.sustainTime && (channel.releaseTime - channel.sustainTime) > RELEASE_TIME_TRIANGLE) {
+        if (time >= channel.sustainTime) {
```
JerwuQu, line 117: I quickly tested.

This change brings back the continous tone bug.

Here's a cart you can use to test the bug: [cart.zip](https://github.com/aduros/wasm4/files/15176401/cart.zip)
With this test cart, only the "Short attack" should be popping and all other ones should be smooth.

JerwuQu (changes requested):

runtimes/web/src/apu-worklet.ts
majaha: Ah yes, so it does.

Can I ask what your specific goal with the whole tick thing was anyway? The more I think about it, the more I think that messing with the ADSR envelope using ticks is weird, and possibly overcomplicating things. If the user asks for the envelope to be only one tick long, why would they be surprised when it ends within one tick?

JerwuQu: > If the user asks for the envelope to be only one tick long, why would they be surprised when it ends within one tick?

Put another way: if a user asks for the tone to be one tick long, shouldn't they be surprised if isn't actually one tick long? That doesn't make sense in terms of the abstraction WASM-4 is. There have been bug reports from people being confused about it behaving like that when wanting to make continous tones.

---

More technically (which I plan to go into in a blog post), custom volume/pitch envelopes and effects like vibrato are simply not feasible if you can't rely on `tone` to last the length you say it should last. You can do workarounds by extending each tone to be longer (e.g. 2 or 3 ticks), but then you end up with pretty inaccurate envelopes instead that will overextend in whatever direction they are heading towards, especially if framerates are unstable. In these cases it's better to have it stay stable on the last known volume and pitch instead.

Some other features that are very difficult to do if tone length can't be relied upon are live playback (e.g. virtual piano) with working ADSR, and temporarily muting audio and then unmuting with retained note and envelope state.

The issue I am concerned with is specifically when the `tone` is only one tick long. It should last until the next tick.

Worth noting is all of these issues would disappear if the `update` loop is bound to the audio callback which I plan on giving a shot later, but it's a big effort.

majaha: In the previous version, when the user asked for a note that was the length of one tick, that's exactly what they got: a note that was precisely 1/60th of a second long. I think the problem of confused users may be better serviced with documentation saying that only long or overlapping notes are guaranteed to be continuous.

Phrased another way: There's a lot to be said for taking the approach of "a note is a note is a note", i.e. given a particular tone() call, the sound produced is always exactly the same, in length, volume, and pitch. No stretching or other irregularities can occur.

Are the main problems you have with live playback to do with the pitch gliding? I can think of solutions to other problems you might have, but trying to make the pitch-glides work while making extra-duration tone() calls does present a tricky challenge.

I do also see the value in having the note lengths stay in sync with the game's ticks. Especially as the developer can schedule sound events over 10 seconds into the future, and the tone() API presents the lengths as a number of ticks (I wonder whether the same issue would be brought up if tone lengths were in an integer number of milliseconds, or a float of seconds). I see these two approaches as competing ideas, both with their own pros and cons.

However, my gut feeling is that if a tick-based solution is chosen, then the *whole note* should be tick-based, with each ADSR point waiting for it's tick to come in. The developer could use any of the ADSR points as an important sync point, so we should sync them all. What do you think of that idea? I think I'll try to implement this.
(Another idea is to sync on *every* tick, but that may be overkill.)

> Worth noting is all of these issues would disappear if the update loop is bound to the audio callback which I plan on giving a shot later, but it's a big effort.

About this, I can't see how that's going to be workable. At least, not in the Web runtime, which I know better. You could try though.

majaha:
Alrighty, bit of a behemoth patch here. I made the APU completely tick based, and I also ended up rewriting the frame timing code so that it's much smoother now, whereas before even in perfect conditions it would microstutter. Wasm-4 can also now keep running in a background tab, which is great for the audio experience, and should make netplay much less annoying to work with too. There's also a few other little improvements and clean-ups sprinkled in.

JerwuQu:
Sounds awesome! Can give it a proper look after I'm back from summer vacation, but if it works properly for the cart linked [here](https://github.com/aduros/wasm4/pull/716#discussion_r1586176934) and for music in games currently published on wasm4.org it has my approval!

majaha:
Yup, it should be all good!
I've thrown some builds up on my web server:
http://majaha.com/temp/w4/jrwq
http://majaha.com/temp/w4/journey-to-entorus

Hopefully you can see that it all sounds very smooth and in-time. And that you can pause Journey to Entorus or minimise the browser window and it behaves nicely. Also the falling stars should move in a more regular way, hopefully!

aduros:
@JerwuQu Should we merge this? 👀

majaha:
> @JerwuQu Should we merge this? 👀

Nah, not right now. I think I'm going to have another pass at some of these things, eventually

majaha:
I'm closing this in favour of other pull requests.

PR #717, opened 2024-05-13
rect: fix bounds

JerwuQu:
If x+width goes off-screen in the negative it would corrupt the framebuffer.


aduros:
Nice catch, thanks!

PR #735, opened 2024-06-29
Add font size to constants

ov7a:
Might be useful for text positioning (e.g. centering).


aduros:
Makes sense, thank you!

PR #742, opened 2024-07-21
compile with --threads:off Update cart.nimble

0riginaln0:
When compiling the cart, an error occurs:
Threads not implemented for os:any. Please compile with --threads:off. stack trace: (most recent call last)

Added --threads:off flag, for successful compilation of the cart


aduros:
Thanks!

PR #745, opened 2024-07-29
Fixing the size of Rust binaries

Ztry8:
The size before the fix was `496` bytes! (without wasm-opt)
The size is now `286` bytes! (also without wasm-opt)


Ztry8:
Errors with `Vec` please do not merge yet!
in the works!

Ztry8:
Fixed vecs now! [See also](https://docs.rust-embedded.org/book/intro/no-std.html#overview)
A snake tutorial fix is already in the works!

Ztry8:
In summary, the following changes were made:
- The size without wasm opt was reduced from `496` to `293`, which represents a `60%` reduction!
- The dependencies were removed and, upon testing, appear to be functioning correctly without them!
- The snake game tutorial has been updated!

Thank you for this amazing and very cool library!

Ztry8:
The size did not change after using wasm-opt, maybe it is no longer needed and need to update the [documentation here](https://wasm4.org/docs/tutorials/snake/publish-your-game)

0riginaln0:
Nice! The cart size has been reduced from `41960` to `20592`

PR #746, opened 2024-07-31
fix retroarch core running too fast under certain conditions

joyrider3774:
I had an issue on my rg35xxsp with latest muos cfw test build. Where retroarch reported games running at about 120 fps with vsync on this made the games run twice as fast. i was able to fix it when testing by making the destination buffer a global and only calling video_cb once at the end of retro_run this makes sure that `video_cb(destframe, 160, 160, ...` is only ever called really once per frame. My guess is that it was potentially being called multiple times per frame in my case.

Now when i ran the core on my rg35xxsp it showed correctly 60 fps in the fps counter and games did not run twice as fast anymore.

It was also really noticeable while running the core with vsync off, on my rg35xxsp retroarch was reporting 1200 fps and the games were running equally fast then but with this change vsync on or off it was running always at 60 fps

I had heared on discord this may have been a known issue or was reported at least once. I do not know if this fix is valid but at least for me it fixed it

### DIFF
--- a/runtimes/native/src/backend/main_libretro.c
+++ b/runtimes/native/src/backend/main_libretro.c
@@ -1,3 +1,4 @@
+
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


joyrider3774:
hmmm the PR is bogus what actually fixes it for me is setting audio to normal instead of callback and it just happens that i had made that change as well while testing around but it only applied after a restart

Issue #750, opened 2024-08-10
native runtime doesn't seem to have a proper memory bounds check


yamt:
looking at the implementations of host functions (blit etc)
they seem only validating the given pointer is pointing inside the wasm memory.
it isn't enough for functions which access more than one byte. (ie. most of them.)

Issue #709, opened 2024-04-18
[native runtime] BSOD on pointer buffer overruns and invalid memory access


JerwuQu:
See this comment for previous discussion https://github.com/aduros/wasm4/pull/599#pullrequestreview-1203505916.

Functions that take pointers in the C runtime should check that they are valid rather than silently trim any out-of-bounds access. If such cases are detected, BSOD so the user can fix the issue.

JerwuQu:
Fixed!

aduros:
Fixed by #751

PR #751, opened 2024-08-14
native runtime: implement memory bounds check

yamt:
Fixes:
https://github.com/aduros/wasm4/issues/750
https://github.com/aduros/wasm4/issues/709

### DIFF
--- a/runtimes/native/src/runtime.c
+++ b/runtimes/native/src/runtime.c
@@ -67,7 +67,7 @@ static void bounds_check(const void *sp, size_t sz)
     const void *memory_sp = (const void *)memory;
     const void *memory_ep = (const uint8_t *)memory_sp + (1 << 16);
     const void *ep = (const uint8_t *)sp + sz;
-    if (ep <= sp || sp < memory_sp || memory_ep < ep) {
+    if (ep < sp || sp < memory_sp || memory_ep < ep) {
         out_of_bounds_access();
     }
 }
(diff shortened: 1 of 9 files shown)


runtimes/native/src/runtime.c
```diff
@@ -83,6 +129,9 @@ void w4_runtimeBlitSub (const uint8_t* sprite, int x, int y, int width, int heig
     bool flipX = (flags & 2);
     bool flipY = (flags & 4);
     bool rotate = (flags & 8);
+    uint32_t bpp = (int)bpp2 + 1;
+    uint32_t nbits = mul_u32_with_overflow_check(mul_u32_with_overflow_check(width, height), bpp);
+    bounds_check(sprite, nbits / 8);
```
aduros, line 134: I like the bounds check here, but do we necessarily need to do an overflow checked multiply? I think we can live with possible overflows, as long as it doesn't allow out of bounds access somehow?

yamt: to implement bounds check that way, we need to make this module (runtime.c) know the implementation details of w4_framebufferBlit. or, implement bounds check in w4_framebufferBlit itself.

IMO, it's simpler to perform overflow checks here.

aduros: Ah, I see. Passing huge values for width/height here that result in an overflow here would fool the bounds check and result in OOB in w4_framebufferBlit. Nice catch!

Issue #497, opened 2022-08-14
Add ability for `png2src` to handle any width/size


JerwuQu:
By padding the last byte, any size would work.

It seems like the `blit` function is already able to handle data with widths not divisible by 4 or 8, but I have no clue about `blitSub`.

PR #760, opened 2024-09-07
add png2src unit tests

lukki15:
relates to #497

Because the math of handling any PNG width is a bit tricky, I created this unit tests.

I used the Jest unit testing framework and activated the coverage report, which says the tests cover 98.91% of lines.
At this moment I verify the code by having sprites of size 7x7 and 8x8 with one and two BPPs.


aduros:
Thank you!! We're overdue for adding some tests 😄

PR #761, opened 2024-09-13
Enable bulk memory for C template

yamt:


yamt:
see https://github.com/aduros/wasm4/pull/762 for the ci failure.

PR #763, opened 2024-09-14
fix: transparent oval outline

### LINKED ISSUE #324: [Bug] oval() does not support transparent outline
The following example (in Rust) draws an oval and a rectangle (as expected), but the oval has a dark outline:
```rust
#[no_mangle]
fn update() {
    unsafe { *DRAW_COLORS = 2 }
    oval(40, 20, 80, 40);

    unsafe { *DRAW_COLORS = 2 }
    rect(40, 80, 80, 40);
}
```
I would expect this code (and the equivalent `DRAW_COLORS = 0x02`) to draw a filled oval with no outline.

This seems to be either a bug or [the documentation](https://wasm4.org/docs/reference/functions#oval-x-y-width-height) could be more explicit, especially because [the hello world example](https://wasm4.org/docs/getting-started/hello-world#accessing-memory) also uses `DRAW_COLORS = 2`.

### Version

`w4 --version` is 2.2.0

### Workaround
Replacing `2` with `0x22` works as it draws the outline in the same color as the rest of the oval.

peter-jerry-ye:
Closes #324 

Before: 
截屏2024-09-14 10 40 19

After:
截屏2024-09-14 10 40 48

### DIFF
--- a/runtimes/native/src/framebuffer.c
+++ b/runtimes/native/src/framebuffer.c
@@ -426,10 +426,12 @@ void w4_framebufferOval (int x, int y, int width, int height) {
     b1 = 8 * b2;
 
     do {
-        drawPointUnclipped(strokeColor, east, north); /*   I. Quadrant     */
-        drawPointUnclipped(strokeColor, west, north); /*   II. Quadrant    */
-        drawPointUnclipped(strokeColor, west, south); /*   III. Quadrant   */
-        drawPointUnclipped(strokeColor, east, south); /*   IV. Quadrant    */
+        if (dc1 != 0) {
+            drawPointUnclipped(strokeColor, east, north); /*   I. Quadrant     */
+            drawPointUnclipped(strokeColor, west, north); /*   II. Quadrant    */
+            drawPointUnclipped(strokeColor, west, south); /*   III. Quadrant   */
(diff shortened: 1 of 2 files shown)


majaha:
I think the idea here is good, but it doesn't look quite right: your second circle is skinnier than the first one.

There are also several pre-existing weird qualities of the code that I'm going to point out (Not your fault, I just want to call them out as I see them). This code could probably do with a cleanup anyway

majaha:
From above:
```
 if (dc1 == 0xf) {
        return;
}
```

Secret feature to skip the draw? This code seems to be really old, I think it's probably unintentional.

```
uint8_t strokeColor = (dc1 - 1) & 0x3;
uint8_t fillColor = (dc0 - 1) & 0x3;
```
This underflows, which was what was causing the dark colour you were seeing. Kinda weird behaviour.

Probably this code needs some real scrutiny and re-writing, but it's complex and people are afraid to touch it so it hasn't gotten any.

peter-jerry-ye:
@majaha You are right about the size. I gave it a quick fix by aligning the color to the former one. Though I do think it may need a proper rewrite.

截屏2024-09-14 15 27 32

PR #762, opened 2024-09-13
CI: Update upload-artifact and download-artifact

yamt:
Fixes the following error:
```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/
```

PR #764, opened 2024-09-15
update a game lad2024

yamt:
* add several stages

* improve animation

* reduce idle cpu usage


yamt:
see https://github.com/aduros/wasm4/pull/762 for the ci failure.

PR #772, opened 2024-10-15
Fix typos

sunarch:
Hi,

I've found a couple of typos while reading the docs, so here are some edits.

This is a terrific project, I'm happy to have found it:)


aduros:
Nice catch, thanks!

sunarch:
You're welcome:)

Issue #779, opened 2024-12-01
Bug with WS global patching causing netplay black screen


CanyonTurtle:
## Bug description
Currently, netplay URLs do not load on desktop web and mobile web.

## Reproduce
On any computer browser (e.g. chrome), visit a wasm4 game (e.g. https://wasm4.org/play/kittygame) and press `enter->copy netplay URL`. Now open another tab and paste. Observe a black screen. Open developer console and see the error:

```
Uncaught (in promise) TypeError: set WebAssembly.Global.value): Can't set the value of an immutable global.
    at Ct.write (wasm4.js:50:9105)
    at RTCDataChannel.<anonymous> (wasm4.js:50:20165)
```

## Fix exploration

Now observe that if you use devtool source overrides (https://stackoverflow.com/a/61377922), comment out the problematic line (wasm4.js line 50), save, and reload, netplay functions as normal.

This appears to be caused by #715 with [this specific change](https://github.com/aduros/wasm4/pull/715/files#diff-5350282c31672fdefd1b39369d867d93f50436c107f1779ed80f86dec486f6acR41). @JerwuQu any ideas for a fix? I could look myself but I don't want to jump in and break global patching, as I don't know how that part of the codebase works just yet.

JerwuQu:
Seems to be working now! Can this be closed?

PR #780, opened 2024-12-03
 web: safely handle immutable globals #779

JerwuQu:
See #779

Though this looks a bit ugly, it beats the alternative.

**Explanation:**
JavaScript doesn't support checking directly whether a `Global` object is mutable or not, and parsing enough WASM to determine if each global is mutable [requires being able to parse WASM expressions](https://webassembly.github.io/spec/core/syntax/modules.html#syntax-global), meaning all possible opcodes permitted in initialization and their parameters. I gave it a quick attempt still, but the amount of opcodes that you have support depends on WASM extensions (e.g. globals using reference types and vector types add more opcodes you need to be able to handle in expressions), and simply adding the ones available today doesn't feel future-proof.
I'm also fine with reverting the global patching completely, though it would be a bit unfortunate.
Another alternative is using a proper WASM parsing library, but that increases bundle size instead, and would still need updating if a new WASM extension is ever made.

The `DEV_NETPLAY` addition is for it to use the `http://localhost:3000/?netplay=xyz` URL rather than wasm4.org when enabled, which eliminated some domain issues I was having otherwise. I figured I'll look into that issue in a separate PR.

### DIFF
--- a/runtimes/web/src/netplay/index.ts
+++ b/runtimes/web/src/netplay/index.ts
@@ -252,10 +252,10 @@ export class Netplay {
         const localPeerId = await this.peerMgr.localPeerId;
 
         const loc = window.location;
-        if (loc.protocol == "file:"
+        if (!DEV_NETPLAY && (loc.protocol == "file:"
             || loc.hostname == "localhost"
             || loc.hostname == "127.0.0.1"
-            || loc.hostname == "wasm4.org") {
+            || loc.hostname == "wasm4.org")) {
             return `https://wasm4.org/netplay/#${localPeerId}`;
         }
         const url = new URL(loc.href);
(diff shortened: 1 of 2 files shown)


CanyonTurtle (approved):
Thanks @JerwuQu! 

What do you guage the impact of immutable globals silently failing is? Is it worth debug.erroring to this Pr/issue chain for early adopters writing GC games?

JerwuQu:
This alone won't affect GC games, and I think logging these exceptions won't be useful since they'd happen every tick.

I haven't read up on the WASM GC proposal in detail but afaik GC objects are stored outside of the main WASM memory and it would therefore be hard to get GC games working with netplay at all (unless JavaScript has a way to fully manipulate the GC heap, which is pretty unlikely).

Unless I'm mistaken about the above *or* we go all in on very complex WASM patching, we would likely need to change how netplay works and freeze the whole game at start and then have a "start multiplayer" button which locks in all current players and prevents any new from joining. We'd also have to remove rollback netcode completely since that depends on being able to swap memories. Not desirable!

**In short:** GC games will likely be broken in netplay, but it won't be because of globals.

aduros:
Thanks!

I think it's a fine solution, and agree we probably don't want to get crazy with opcode parsing.

PR #787, opened 2024-12-12
return png2src error to caller

### LINKED ISSUE #786: png2src return value does not report errors to the caller
If png2src encounters an error, it does not report it via its return value:

```bash
$ w4 png2src --zig lib/test/data/smile_8x8_5colors.png
Error processing lib/test/data/smile_8x8_5colors.png: 
Too many colors: maximum is 4. The previous colors were:
- (R: 0, G: 0, B: 0, A: 255); first seen at (0, 0)
- (R: 255, G: 255, B: 255, A: 255); first seen at (2, 0)
- (R: 0, G: 255, B: 0, A: 255); first seen at (2, 2)
- (R: 255, G: 0, B: 0, A: 255); first seen at (2, 5)
The first occurrence of another color is at (3, 6) and has the value of (R: 0, G: 0, B: 255, A: 255)
```

Expected behavior:
```bash
$ echo $?
1
```

Actual behavior:
```bash
$ echo $?
0
```

peppergrayxyz:
fix #786

### DIFF
--- a/cli/lib/test/png2src.test.js
+++ b/cli/lib/test/png2src.test.js
@@ -165,11 +165,11 @@ describe("PNG2SRC", () => {
     });
 
     it("should abort and output an error message for invalid image paths", () => {
-      runAll(["/invalid/path"], {
+      expect(() => 
+        runAll(["/invalid/path"], {
         lang: "rust",
         output: "-",
-      });
-      expect(console.error).toHaveBeenCalledWith("Error processing /invalid/path: ENOENT: no such file or directory, open '/invalid/path'");
+      })).toThrowError(/Error processing \/invalid\/path: ENOENT: no such file or directory, open '\/invalid\/path'/);
     });
 
     it("should console log one sprite", () => {
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -173,7 +173,12 @@ program.command("png2src <images...>")
         if (!opts.template) {
             opts.lang = requireLang(opts);
         }
-        png2src.runAll(images, opts);
+        try {
+            png2src.runAll(images, opts);
+        } catch (error) {
+            console.error(error.message);
+            process.exit(1);
+        }
     });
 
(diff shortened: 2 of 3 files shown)


cli/lib/test/png2src.test.js
```diff
@@ -165,11 +165,11 @@ describe("PNG2SRC", () => {
...
 
     it("should abort and output an error message for invalid image paths", () => {
-      runAll(["/invalid/path"], {
+      expect(() => 
+        runAll(["/invalid/path"], {
         lang: "rust",
         output: "-",
-      });
-      expect(console.error).toHaveBeenCalledWith("Error processing /invalid/path: ENOENT: no such file or directory, open '/invalid/path'");
+      })).toThrowError(/Error processing \/invalid\/path: ENOENT: no such file or directory, open '\/invalid\/path'/);
```
matm, line 172: @peppergrayxyz I think it could be worth asserting the process exits with code error 1 with something like
```
const mockExit = jest.spyOn(process, 'exit').mockImplementation(() => {});
// expect with runAll() call here then
expect(mockExit).toHaveBeenCalledWith(1);
```
WDYT?

peppergrayxyz: I tried to fit it into the existing logic and `run` in `png2src.js` does not return a value either, but only throws an exception, so I though `runAll` should behave the same way. Also, the other modules don't return a value either, but throw exceptions. So I think an exception is the most consistent thing to do here.

If we want to check for `return 1` I think we should do it in the test for `cli.js` and then also for all such cases, but there are no tests for that file (yet).

aduros:
Thanks!

cli/lib/test/png2src.test.js
matm: Okay, I don't know this tool yet, you are perfectly right. Thanks for your detailed answer.

PR #793, opened 2024-12-20
Allow running web runtime in background windows

majaha:
This patch makes update timing more regular in various environments, including when the browser window is in the background. This allows audio and netplay to continue if the user e.g. minimises the browser window. This is achieved by switching to a setTimeout() based timing solution if the observed requestAnimationFrame rate is not a multiple of 60.

Also included are a few other relevant fixes and improvements.

### DIFF
--- a/cli/lib/server.js
+++ b/cli/lib/server.js
@@ -41,7 +41,17 @@ async function start (cartFile, opts) {
     }
 
     // Serve the WASM-4 developer runtime.
-    app.use(express.static(path.resolve(__dirname, "../assets/runtime/developer-build")));
+    app.use(express.static(
+        path.resolve(__dirname, "../assets/runtime/developer-build"),
+        {
+            setHeaders: function (res, path, stat) {
+                // These COOP and COEP headers allow us to get high-precision time.
+                // https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#security_requirements
+                res.set("Cross-Origin-Opener-Policy", "same-origin");
+                res.set("Cross-Origin-Embedder-Policy", "require-corp");
+            }
+        }
+    ));
 
 
     const first_port = opts.port;
--- a/devtools/web/package-lock.json
+++ b/devtools/web/package-lock.json
@@ -14,7 +14,6 @@
         "lodash-es": "^4.17.21"
       },
       "devDependencies": {
-        "@types/node": "^20.11.16",
         "prettier": "3.2.4",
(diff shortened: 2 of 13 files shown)


aduros (changes requested):
This is really clever, thanks!

I think we should only keep the game running in the background when netplay is active as a special case. By default we should continue to respect requestAnimationFrame's preference to pause when the browser is in the background, to avoid hogging the system unnecessarily. What do you think?

Please also remove this bit from the docs now that this is resolved: https://github.com/aduros/wasm4/blob/77352e040b78c81c261b7a6862e4a34ec468127c/site/src/components/NetplayCart.js#L42-L43

majaha:
> I think we should only keep the game running in the background when netplay is active as a special case. By default we should continue to respect requestAnimationFrame's preference to pause when the browser is in the background, to avoid hogging the system unnecessarily. What do you think?

I disagree, I think we should rather be respecting the user's wishes. If the user wants to continue to listen to the music in the background, they should be able to do so. It's better that they have the choice to pause the game manually if they would like to.

Halting updates like that also interacts poorly with the audio system, especially tick-based audio. Long notes keep playing, and then are missing when you tab back to the game, it sounds pretty clunky and bad. (~~Also, on Firefox rAF callbacks continue at a slower rate in the background, meaning the game isn't really paused. You can put a game in the background, and then over 15 minutes later it will randomly play note.~~ This is half-true, but Firefox seems to turn this behaviour off when there's an active audio context. It's still worth considering though.)

Allowing the game to continue running also makes development and hot-swapping nicer. You can imagine someone composing music or editing sound effects with the runtime running in the background.

I've also included a patch I had lying around that pauses the audio properly when the menu is open. This stops that strange note-holdover effect I mentioned above happening when you open the menu, but it also hints to the browser that it's okay to throttle timers while the menu is open.
We could potentially stop the game from updating at all while the menu is open, but that would require some non-trivial refactoring (making all input event-based so the menu still works. Or adding a rAF only mode that's used when the menu is open would be a hackier way of doing it. It's probably more hassle than it's worth.)

majaha:
If you really think this should be Netplay-only, then I can change it. But the way I see it, YouTube plays in the background, Spotify plays in the background, [Unity Web Games play in the background](https://sodaraptor.itch.io/bubble-dreams-3d), all other web games I've tried will play in the background. It would be weird if WASM-4 games *didn't* play in the background.

PR #796, opened 2024-12-21
Devkit, devtools and developer experience improvements

majaha:
A few patches that:
- Fix an issue with websockets when running vite
- Fix the web runtime fps counter
- Fix the web runtime devtools palette viewer


aduros:
Thanks!

PR #802, opened 2025-03-06
MELONDS

StelingD:
NO COPYRIGHT OR PIRACY INTENDED. ONLY USE GAMES THAT YOU OWN PHYSICALLY.


aduros:
This cart doesn't load and is missing a readme and screenshot.

RobLoach:
I think you've uploaded the libretro core build of melonds. WASM4 doesn't load libretro cores built with emscripten. WASM-4 loads its own wasm cart format.

PR #804, opened 2025-03-14
Updating array indexing single line zig example

0riginaln0:
Indexing BoundedArray with `[ ]` is no longer supported. That's why in the following example `.get()` is used.

image


aduros:
Thanks!

PR #807, opened 2025-05-21
Ubuntu depreciation bump

CanyonTurtle:


CanyonTurtle:
@aduros deploys down, this is fix

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
                       libretro: lib/wasm4_libretro.dylib
 
                     # glibc 2.31
```
yamt, line 20: this comment seems stale after the change.

i guess an older version of ubuntu was preferred here for glibc compatibility reasons.
have you considered ubuntu-22.04?

CanyonTurtle: I haven't. Yeah comment might be stale. I'm away from workstation but would be worth looking at git blame for that. 22.04 is fine too, hadn't put much thought between the two except for just making it as new as possible

aduros:
Thanks @CanyonTurtle, I just switched it to `ubuntu-latest`... not really sure why the glibc version requirement was added, but hopefully it's been fixed on libretro's end.

Issue #810, opened 2025-06-19
Move website to GitHub Actions + Pages


aduros:
The website is currently built and hosted by netlify, which has been ok but there's not much visibility when builds fail and the website stops updating.

Consider moving the website to be built and deployed as the other CI checks on this repo.

Alternatively there might be some kind of github integration for netlify to improve build visibility, not sure.

PR #811, opened 2025-06-20
fix(site): replicate netlify CI step.

sulewicz:
Replicating the Netlify CI build step in the GitHub Action workflow to address https://github.com/aduros/wasm4/issues/810.
This way any site building failures will be visible in future PRs.
Kind of naive approach but does the job:

image

image


aduros:
Thanks! This is a great first step.

PR #822, opened 2025-07-31
Use gflw as default for the window backend

Tomcat-42:
This pull request defaults the native window backend to GLFW. The current minifb/X11 build, while functional on Wayland via XWayland, has some issues (namely the wayland fractional scale fiasco).

By building with glfw it uses the native wayland API on wayland and the Xorg API on X11. I've tested successfully in the following platforms:

Fedora Rawhide (43 pre-release) [pure wayland]
openSUSE LEAP 15.6 [x11]
Ubuntu 24.10 [wayland + xwayland]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,11 +43,12 @@ jobs:
               run: |
                   ldd --version
                   sudo apt update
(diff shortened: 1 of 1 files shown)


aduros:
Thanks for this. I originally explored minifb because it's a bit smaller/lighter than glfw, but compatibility is more important.

Maybe [sokol_app.h](https://github.com/floooh/sokol) is worth looking into at some point...

PR #827, opened 2025-08-04
native runtime: fix a NULL dereference

yamt:
bail out gracefully when we failed to open argv[0]. note that argv[0] is not necessarily an open-able path.


aduros:
Thanks!

PR #831, opened 2025-08-23
Update Rust template to 2024 edition

wischi-chr:
This PR updates the Rust template to the 2024 edition.

### add explicit unsafe
Some things (that technically always were "unsafe" are now required to be explicitly marked `unsafe` (like `no_mangle` attribute and `extern "C"` functions).

### remove shared ref to static mut
In `alloc.rs` the old code ran into the [Disallow references to static mut](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html) lint (`deny` since 2024 edition) because `.as_ptr()` implicitly took a shared reference to the static mut u8 array. I'm not an unsafe expert but I guess the old use was fine, but there is actually a way to get that pointer without unsafe, by using `&raw` and pointer casting - which is all fine in safe rust because we never turn it into a reference or read from it.

## testing
I'm pretty new to wasm4 (just found it and I think it's awesome) so I just ran `w4 watch` on the new template and that still worked as expected.

### Update 1
Originally I also updates `buddy-alloc` to `0.6.0` but it looks like this causes some memory issues. I have no idea what's to blame. If some unsafe code in Rust is wrong, or something inside buddy-alloc or in the `extern "C"` functions but reverting back to 0.4.1 seems to have fixed the issue.

### DIFF
--- a/cli/assets/templates/rust/Cargo.toml
+++ b/cli/assets/templates/rust/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cart"
 version = "0.1.0"
-edition = "2021"
+edition = "2024"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [lib]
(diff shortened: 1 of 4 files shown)


aduros:
Thanks!

PR #850, opened 2026-01-01
update a game lad2024

yamt:
* a lot of new stages (500+) including the HNY2026 one
* bgm
* animation
* sound effects
* mouse operation
* update a screenshot


aduros:
Reading about how you rediscovered this game, and are still updating it after many years is inspiring! Thank you.

PR #854, opened 2026-03-19
Remove DUB from D template.

Kapendev:
The D template is not working right now. I tried to fix the DUB config, but I didn't find a way to do it.
So, my solution to this was to replace DUB with a shell and batch script. The makefile is still there, so it should continue to work like it used to for people using make.

I also removed the need for the `-allow-undefined` linker flag. To do this, I had to change the wasm.d file a bit.


cyrusmsk:
Dub is the standard package manager in D.
It could be more convenient to use it instead of batch or makefile

PR #856, opened 2026-03-28
Add built-in game store and menu system

WizzardSK:
## Summary
- **Store**: browse and download carts from wasm4.org/carts.json directly in the native runtime
- **Menu**: pause menu (Enter key) with Continue, Store, Reset options
- **No-arg launch**: running `wasm4` without a cart file opens the store directly
- **wasm3 fix**: skip duplicate memory section when memory is already imported (fixes 3 carts: bombfighters, lava-hop, seed-creator-showcase)
- **Store UX**: wrap-around navigation, page up/down, safe loading (failed carts reopen store)

## New files
- `store.c` / `store.h` — store overlay with catalog download, browsing UI, cart download
- `menu.c` / `menu.h` — pause menu overlay

### DIFF
--- a/runtimes/native/CMakeLists.txt
+++ b/runtimes/native/CMakeLists.txt
@@ -120,6 +120,8 @@ endif ()
 add_subdirectory(vendor/cubeb)
 endif ()
 
+find_package(CURL REQUIRED)
+
 file(GLOB COMMON_SOURCES RELATIVE "${CMAKE_SOURCE_DIR}" "src/*.c")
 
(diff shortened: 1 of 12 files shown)


aduros:
Thanks for the PR, but this introduces too much complexity IMO. The native runtime should be small and simple, with no extraneous dependencies, to make long-term maintenance easier.

Would you be interested in porting this to the web runtime? We allow more features in the web runtime, and it would probably be easier to implement (there's already a menu/UI system there for example)

PR #859, opened 2026-05-10
Add Mandelbrot

LukeGrahamLandry:


aduros:
Thanks! When can we get official Franca support? ;)

PR #861, opened 2026-05-17
add NAIRDEMAKE, a simple fighting game with only one attack.

miguelgss:
Source of the project: https://codeberg.org/MiguelGSS/NAIR-Demake


aduros:
Nice! Looks like the perfect netplay fighting game :)

PR #863, opened 2026-05-31
annoyingrobots

sulewicz:
Adding a new game


aduros:
Thanks! This is pretty fun

PR #867, opened 2026-06-16
Make indexes/indices consistent

a-warlock:
I noticed a handful of places in your docs that referred to "indexes" rather than "indices" (where in another handful of places, it was correct).
I've made this consistent, outside of libretro.h from RetroArch.


aduros:
Thanks!

PR #868, opened 2026-06-27
libretro: When reseting, clean up the entire wasm instance

RobLoach:
Found a small potential issue with reseting the libretro core. Since it doesn't clear out the instance, running runtimeInit() would overwrite it.


aduros:
Thanks Rob 🙏

PR #884, opened 2026-08-22
Fix Web runtime rounding up disk length to a multiple of four

majaha:
To save disk data from diskr() and diskw(), the web runtime uses a scheme called z85 which encodes the binary data as base 85. However, z85 [only supports byte lengths that are a multiple of four](https://rfc.zeromq.org/spec/32/#:~:text=The%20binary%20frame%20SHALL%20have%20a%20length%20that%20is%20divisible%20by%204). This causes weird behaviour visible to carts where the disk changes length when you restart the web runtime (but stays the same before a restart). This behaviour differs from the native runtime.

I fixed this by augmenting the z85 implementation with the ability to represent 1, 2, and 3 byte chunks.
There are tests in the intermediary commits that demonstrate this.

Also: some tidy-up on the C runtime.


aduros:
Thanks, this looks great! I've handled this on other projects by storing the original byte length separately. These format changes 100% compatible with the previous format right?

I wouldn't mind having the tests in a separate PR, if you ever feel like wiring up the infra for that.

majaha:
Yes, this is 100% compatible with old save files, sorry I meant to mention that.

PR #885, opened 2026-08-28
Melancholy: cart 3

De-Alchmst:
Once again, I felt like continuing my little art project and once again, it was in WASM-4.


aduros:
I just want to say reading the source has been a treat, especially compared to the gobs of boring LLM output we wade through every day. So thanks for that!

PR #889, opened 2026-09-02
Allow filtering experiments from the list of games

haykam821:
Since games and experiments offer a vastly different experience, I think allowing them to be separated in the list makes sense. Currently, this pull request doesn't classify any cartridges as experiments. Both types of cartridges are selected by default, but I wonder if only showing games by default makes sense.


aduros:
I agree it would be good to have better filtering/discoverability on the site, but I'm not sure about the approach.

How about a general tagging system? "experiment" could be one of the tags, in addition to eg: "puzzle", "multiplayer", "action", "3d", etc.

PR #891, opened 2026-09-04
ci: add a webOS build for the libretro core

WizzardSK:
The libretro CI here covers desktops, Android, iOS/tvOS, six consoles and Emscripten, but not webOS — LG TVs run RetroArch too, and the core needs nothing special to get there.

The `webos-cmake` template is used as-is: `.core-defs` already points `CMAKE_SOURCE_ROOT` at `runtimes/native` and passes `-DLIBRETRO=ON`, which is all the job needs.

Verified by replaying it against the same buildroot SDK the build image bakes in: it produces `wasm4_libretro.so`, `ELF 32-bit LSB shared object, ARM, EABI5`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -101,6 +101,10 @@ include:
     file: '/tvos-cmake.yml'
 
   ################################## MISC ####################################
+  # webOS (LGTV)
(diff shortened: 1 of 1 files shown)


aduros:
Thanks!

